PR #85, opened 2026-06-10
fix(opencode): support for opencode and opencode ACP

### LINKED ISSUE #23: Feature request: support OpenCode as a first-class coding CLI
Hi,

Would you consider adding **OpenCode** as a supported coding CLI for **Liza**?

## Why I'm asking

I'm interested in using Liza, but my current day-to-day setup is **OpenCode** with a **GitHub Copilot** subscription. OpenCode already supports project instruction files via `AGENTS.md`, has `CLAUDE.md` compatibility, and can load additional instruction files through `opencode.json`, so it feels close to Liza's contract-activation model.

## Current blocker

From the docs, Liza currently supports **Claude Code, Codex, Kimi, Mistral, and Gemini**. I don't see an OpenCode path in the supported CLI list or in the setup/init flags.

## Request

The goal is simply to be able to run **Liza on top of OpenCode** so existing OpenCode users can keep their current workflow while using Liza's contract and orchestration.

I'm **not** asking for project-management or GitHub workflow integration here — only support for OpenCode as a coding CLI/runtime.

A smaller first step would already be valuable:

- pairing-mode support first
- MAS support later if needed

## Questions

1. Would OpenCode support fit Liza's scope?
2. If yes, what are the minimum requirements a CLI has to satisfy to be wrapped by Liza?
3. Would you prefer:
   - native OpenCode support,
   - a compatibility mode,
   - or documentation explaining why OpenCode is not a fit today?

If the answer is "not currently planned," a short explanation of the blocker would still be very helpful.

Thanks — this would make Liza much easier to adopt for users who already work in OpenCode.

Additionally OpenCode allready supports a lot of LLM Providers. So supporting OpenCode allows a broader compatibility.

LivioGama:
## Summary
- add OpenCode activation artifacts for Liza workspaces
- install a managed `.opencode/tools/exec.ts` compatibility tool
- document OpenCode/Groq model guidance and bridge behavior
- harden successful-exit no-progress detection to ignore provider output churn while preserving task/worktree progress signals
- bound OpenCode exec output accumulation and document the shell trust boundary
- remove the duplicate archived `specs/vision.md` copy and repair the remaining internal Vision link

Closes #23

## Validation
- `rtk git diff --check`
- `rtk env GOMODCACHE=/tmp/liza-gomodcache GOCACHE=/tmp/liza-gocache go test -race ./internal/embedded ./internal/commands ./internal/agent ./cmd/liza`
- `rtk env GOMODCACHE=/tmp/liza-gomodcache GOCACHE=/tmp/liza-gocache make test`

## Known local hook issue
- `rtk pre-commit run --files ...` still fails in `jscpd` on existing repository-wide duplicates under `.git/hooks`, `.git/lost-found`, README/spec docs, `_local_backup`, and other baseline paths; all non-`jscpd` hooks passed.


tangi-vass:
@LivioGama Do you want to include https://github.com/liza-mas/liza/issues/23?

tangi-vass:
Review: Standard — Request Changes

[blocker] `internal/agent/supervisor.go:546` — OpenCode success-loop detection replaces the task/worktree progress signature with provider output whenever output is nonempty.

Why it matters: `RunSupervisor` reads the post-run task/worktree snapshot, but `successfulTurnProgressSignature` discards it for `opencode`/`opencode-acp`. That can falsely block a valid task if OpenCode makes real worktree progress while emitting the same JSON/output as a prior turn. It can also miss no-progress loops when output changes but task/worktree state does not. This weakens the invariant this guard is meant to protect: blocking infinite loops without blocking real progress.

Suggestion: make state/worktree progress the primary signature for every CLI. If OpenCode output is useful, track it as a separate repeated-output signal or append it without replacing the snapshot. Add tests for both cases: same output with changed worktree must not block; changed output with unchanged task/worktree should still count as no progress if that is the intended guard.

[suggestion] `internal/embedded/opencode-tools/exec.ts:67` — The output limit is applied only after stdout/stderr are fully buffered.

A command that emits very large output can grow `stdout`/`stderr` unbounded before `truncateOutput` runs. Consider capping during accumulation, or killing/ignoring additional output once the limit is reached.

[suggestion] `support-docs/USAGE_MULTI_AGENTS.md:213` — This CLI list still omits `opencode` and `opencode-acp`, despite the updated list at lines 160-162.

Overall: OpenCode activation wiring is mostly coherent, and the custom tool shape matches current OpenCode docs for `.opencode/tools/*.ts`. The runtime no-progress guard needs a fix before merge because it can misclassify progress.

Blast Radius: High: agent lifecycle / infinite-loop protection
Confidence: medium-high: reviewed PR metadata, bounded diffs, local source lines, tests, invariants sections, CI status, and OpenCode docs.

internal/agent/supervisor.go
```diff
@@ -541,6 +541,22 @@ func detectLizaCommandContext(output string) string {
...
+	case "opencode", "opencode-acp":
+		trimmed := strings.TrimSpace(output)
+		if trimmed != "" {
+			if len(trimmed) > successfulTurnOutputSignatureLimit {
+				trimmed = trimmed[:successfulTurnOutputSignatureLimit]
+			}
+			return "provider:" + cliName + "\noutput:" + trimmed
+		}
+	}
+	return taskSnapshot
```
tangi-vass, line 562: **[concern]** `successfulTurnProgressSignature` includes raw output for OpenCode, which may cause false positive spin detection resets.

The function uses the first 16KB of OpenCode output as part of the progress signature. If OpenCode produces slightly different output each turn (timestamps, random IDs, different error formatting) while making no actual task/worktree progress, the signature will differ each time, defeating the no-progress spin detection. Other CLIs fall through to `taskSnapshot` which checks actual state changes.

**Why it matters:** The whole point of this PR's supervisor hardening is catching OpenCode's successful-but-spinning behavior. If OpenCode naturally produces varying output, the guard won't fire.

**Suggestion:** Consider falling through to `taskSnapshot` as the default, and only using output-based signatures for the specific case where the task snapshot is empty/unavailable. Or document why output-based detection is intentionally preferred for OpenCode.

internal/agent/progress_watchdog.go
```diff
@@ -156,6 +156,29 @@ func readExecutionProgressSnapshot(projectRoot string, bb *db.Blackboard, taskID
...
+		sig, sigErr := lizagit.New(projectRoot).WorktreeProgressSignature(taskID)
+		if sigErr != nil {
+			worktreeSignature = "error:" + sigErr.Error()
+		} else {
+			worktreeSignature = sig
+		}
+	}
+
+	return exit42TaskProgressSignature(task) + "\nworktree:" + worktreeSignature, true, nil
+}
```
tangi-vass, line 188: **[concern]** `readTaskStateWorktreeProgressSnapshot` duplicates `readExecutionProgressSnapshot` without the assignment/execution-status guard.

The new function is identical to `readExecutionProgressSnapshot` except it omits the `task.AssignedTo == agentID && IsExecutingStatus` check. This means the success-path tracker will consider a task "eligible" even when the task isn't assigned to this agent or isn't in an executing state.

**Why it matters:** If a task gets reassigned mid-turn, the tracker would still accumulate a hit against the wrong agent's count.

**Suggestion:** Either extract the shared worktree-signature logic into a helper and compose both variants, or document why the relaxed eligibility is intentional for the success-no-progress path.

internal/embedded/opencode-tools/exec.ts
```diff
@@ -0,0 +1,90 @@
...
+      typeof args.timeout_ms === "number" && Number.isFinite(args.timeout_ms) && args.timeout_ms > 0
+        ? args.timeout_ms
+        : DEFAULT_TIMEOUT_MS
+
+    return await new Promise<string>((resolve) => {
+      let stdout = ""
+      let stderr = ""
+      let timedOut = false
+
+      const child = spawn(args.cmd, {
```
tangi-vass, line 84: **[suggestion]** `shell: true` with string `cmd` is a command injection surface.

Since this tool runs in the context of an agent that already has full shell access (OpenCode with `--dangerously-skip-permissions`), this isn't a privilege escalation. But if the exec tool is ever exposed to less-trusted contexts, the raw `shell: true` + user string pattern would be exploitable. Worth a brief note in the managed header or tool description.

internal/agent/cli_agent.go
```diff
@@ -399,6 +401,14 @@ func buildCodexArgs(prompt string, useStdin bool, outputsDir string) []string {
 	return args
 }
 
+func buildOpenCodeArgs(prompt string, outputsDir string) []string {
+	args := []string{"run", prompt, "--dangerously-skip-permissions"}
```
tangi-vass, line 408: **[suggestion]** Prompt passed as positional argument to `opencode run`.

For very large Liza prompts (which can be 10KB+), this may hit OS argument length limits (~128KB on Linux, but lower on some systems). Claude and Codex have stdin-based alternatives. The `cliSupportsStdin` function already returns `false` for OpenCode, so this is intentional — just noting the limit.

specs/vision.md
```diff
@@ -0,0 +1,144 @@
+# Liza Vision
+
+> Archived snapshot: this file reflects historical v1 planning scope and may not match current implementation.
+> Current scope lives in `specs/functional/1 - Liza.md` and architecture specs.
```
tangi-vass: **[suggestion]** New file appears to be a historical/archived vision snapshot.

The header says "Archived snapshot" but it's added at `specs/vision.md` which is a fairly prominent location. The current vision lives at `specs/build/0 - Vision.md`. Should this live in `specs/_archive/` to match the existing convention, or is the prominent placement intentional?

internal/commands/init.go
```diff
@@ -55,22 +55,24 @@ type InitParams struct {
 // InitAgentRepoSymlinks maps agent flag names to the repo-root symlink filename.
 // These symlinks point to ~/.liza/CORE.md and enable pairing mode.
 var InitAgentRepoSymlinks = map[string]string{
-	"claude": "CLAUDE.md",
-	"codex":  "AGENTS.md",
-	"gemini": "GEMINI.md",
+	"claude":   "CLAUDE.md",
+	"codex":    "AGENTS.md",
+	"opencode": "AGENTS.md",
```
tangi-vass, line 62: **[suggestion]** Codex and OpenCode share the `AGENTS.md` repo-root symlink target. The `createContractSymlinksForAgents` function handles this gracefully (second pass sees the symlink already exists and skips), and the brownfield test confirms both get independent global fallbacks. This works — worth a brief comment in the map noting the intentional sharing, since it's a non-obvious design choice.

tangi-vass:
Addressed Claude's review comments in `6a743c4d`.

What changed:
- Successful-exit no-progress detection now ignores provider output variation, captures the snapshot while the task is still executing/owned, and includes real worktree progress via `WorktreeProgressSignature`.
- Added regression coverage for claim-churn normalization, actual worktree changes, ownership/status eligibility, and the OpenCode successful no-progress loop.
- Bounded OpenCode exec output accumulation while buffering stdout/stderr and documented the shell trust boundary.
- Added the OpenCode argv-limit note and the intentional shared `AGENTS.md` init comment.
- Updated multi-agent CLI docs for `opencode` / `opencode-acp`.
- Removed duplicate `specs/vision.md` and fixed the remaining internal Vision link to `specs/build/1 - Vision.md`.
- Updated the PR body with `Closes #23`.

Validation:
- `rtk git diff --check`
- `rtk env GOMODCACHE=/tmp/liza-gomodcache GOCACHE=/tmp/liza-gocache go test -race ./internal/embedded ./internal/commands ./internal/agent ./cmd/liza`
- `rtk env GOMODCACHE=/tmp/liza-gomodcache GOCACHE=/tmp/liza-gocache make test`

Known local hook issue: `rtk pre-commit run --files ...` still fails in `jscpd` on existing repository-wide duplicates under `.git/hooks`, `.git/lost-found`, README/spec docs, `_local_backup`, and other baseline paths. All non-`jscpd` hooks passed.

internal/git/worktree.go
```diff
@@ -197,7 +200,64 @@ func (g *Git) WorktreeProgressSignature(taskID string) (string, error) {
...
+			h.Write([]byte(target))
+			h.Write([]byte{0})
+			continue
+		}
+		if !info.Mode().IsRegular() {
+			h.Write([]byte(info.Mode().String()))
+			h.Write([]byte{0})
+			continue
+		}
+		content, readErr := os.ReadFile(fullPath)
```
tangi-vass: [concern] `os.ReadFile(fullPath)` reads the **entire content** of every untracked file into memory to hash it. If an agent creates a large generated file, binary artifact, or dumps a core file in the worktree, this will spike memory. The signature is computed on every progress watchdog tick, amplifying the risk.

**Why it matters:** Progress watchdog runs on a recurring cadence. A single large untracked file (build artifact, test fixture, model weight) turns a lightweight polling operation into a multi-hundred-MB allocation per tick.

**Suggestion:** Stream through the hasher with `io.Copy(h, file)` instead of `os.ReadFile`. Or hash only file size + mtime as a cheaper proxy (the tracked diff path already gets content via `git diff --binary`). Consider capping individual file reads or skipping files above a threshold.

internal/git/worktree.go
```diff
@@ -197,7 +200,64 @@ func (g *Git) WorktreeProgressSignature(taskID string) (string, error) {
...
+	h.Write([]byte("tracked\x00"))
+	h.Write(diff)
+
+	untracked, err := gitenv.CombinedOutput(worktreePath, "ls-files", "-z", "--others", "--exclude-standard")
+	if err != nil {
+		return "", fmt.Errorf("git ls-files failed: %w\nOutput: %s", err, untracked)
+	}
+	h.Write([]byte("\x00untracked\x00"))
+	h.Write(untracked)
+	for _, relPath := range strings.Split(strings.TrimSuffix(string(untracked), "\x00"), "\x00") {
```
tangi-vass: [suggestion] When `ls-files` returns empty output (no untracked files), `strings.Split("", "\x00")` produces `[""]` — the loop runs once with `relPath == ""` and the `continue` catches it, so it's correct. But the empty-input edge would be cleaner with an early return before the split:
```go
if len(untracked) == 0 {
    return fmt.Sprintf("%x", h.Sum(nil)), nil
}
```

internal/git/worktree.go
```diff
@@ -197,7 +200,64 @@ func (g *Git) WorktreeProgressSignature(taskID string) (string, error) {
...
+	if err != nil {
+		return "", err
+	}
+	return head + "\n" + status + "\ncontent:" + contentSig, nil
+}
+
+func (g *Git) worktreeDirtyContentSignature(worktreePath string) (string, error) {
+	h := sha256.New()
+
+	diff, err := gitenv.CombinedOutput(worktreePath, "diff", "--no-ext-diff", "--binary", "HEAD", "--")
```
tangi-vass: [question] `git diff --binary HEAD` on every watchdog tick adds ~2 git process spawns (diff + ls-files) per tick per active agent. Is this measurably heavier than the previous HEAD+status pair? For large worktrees with many modified files, this could add latency to the polling loop.

tangi-vass (commented):
Review: Standard — Comment (scoped to latest 2 commits: `042ba1bb`, `ee0c743a`)

Blockers: None
Concerns: 1 (unbounded file read in progress signature)
Suggestions: 1
Questions: 1

Overall: Two focused hardening commits. The exec timeout escalation is clean. The progress signature enhancement correctly closes a gap where agents could edit files without commits and appear "stuck." The main concern is that `os.ReadFile` in the untracked-file hashing path is unbounded and runs on a recurring timer — streaming or capping would be safer.
Blast Radius: Medium — affects progress watchdog behavior for all active agents
Confidence: High — read all changed files and their callers

🤖 Generated with [Claude Code](https://claude.com/claude-code)

tangi-vass:
Review: Standard - Comment

I found two concerns in the latest two commits. I did not run tests.

`internal/embedded/opencode-tools/exec.ts:68` - Timeout kill only targets the immediate shell process.

Why it matters: with `shell: true`, a timed-out command can leave child processes running after Liza reports `timed_out: true`, especially for commands that spawn workers or background children. That can keep mutating files or burning CPU after the supervisor believes the command is done.

Suggestion: spawn a process group where supported and kill the group on timeout, with a behavioral test that starts a child process and verifies it is gone after timeout.

`internal/git/worktree.go:213` - Progress signatures now capture full binary diffs and read every untracked file into memory.

Why it matters: `WorktreeProgressSignature` runs from the agent progress watchdog on polling intervals. A task that creates large generated files, model outputs, archives, or binary diffs can make the watchdog expensive or stall it, which weakens the timeout guard it is meant to support.

Suggestion: make the signature bounded and streaming: avoid `git diff --binary` through `CombinedOutput`, avoid `os.ReadFile` for untracked content, and include size/mtime plus bounded content hashing or streamed hashes with limits.

No blockers found.

tangi-vass (approved):

PR #88, opened 2026-06-11
Add opt-in Liza update checks

LivioGama:
## Summary
- add opt-in `--check-update` support before normal CLI execution
- support `stable` release-binary updates and `main` source-build updates via `--update-channel`
- persist declined-check disable preference in `~/.liza/update.json`, while explicit `--check-update` enables checks again
- re-run the original command after successful update with an internal skip marker to avoid loops

## Verification
- `go test ./internal/updater ./cmd/liza`
- `go build -o /tmp/liza-check-update-final ./cmd/liza && /tmp/liza-check-update-final version`
- TTY stable-channel harness: downloaded a local fake release binary, replaced the binary, and re-execed the original command
- TTY main-channel harness: fake-cloned repo, checked out the exact main commit, ran `make install`, and re-execed the original command
- TTY decline flow: declined update, accepted disable proposal, wrote `~/.liza/update.json`, then reran with `--check-update` and confirmed the explicit flag enabled checking again

## Known unrelated failures
`go test ./...` still fails on existing unrelated issues:
- embedded artifact drift in `internal/embedded`
- global `core.excludesFile=/Users/livio/.gitignore_global` affecting worktree exclude/indexing tests

### DIFF
--- a/cmd/liza/main.go
+++ b/cmd/liza/main.go
@@ -231,6 +231,11 @@ func init() {
 	// Global flags
 	rootCmd.PersistentFlags().BoolP("verbose", "v", false, "verbose output")
 	rootCmd.PersistentFlags().StringP("project-root", "C", "", "Liza project root for state commands")
+	// Note: --check-update and --update-channel are registered here for Cobra help visibility,
+	// but the updater package manually pre-parses these flags before Cobra command execution.
+	// This allows update checks to run before the main command, while still showing these flags
+	// in --help output. The manual parsing in internal/updater implements stop-at-double-dash
+	// and last-flag-wins semantics to match pflag/Cobra behavior for these specific flags.
 	rootCmd.PersistentFlags().Bool("check-update", false, "check for a Liza update before running")
 	rootCmd.PersistentFlags().String("update-channel", "stable", "update check channel: stable or main")
 }
@@ -266,8 +271,15 @@ func main() {
 		CurrentCommit:  GitCommit,
 		IsInteractive:  interactive.IsInteractive,
 	}); err != nil {
+		// FatalError represents invalid CLI input/config that should exit before command execution
+		var fatalErr *updater.FatalError
+		if errors.As(err, &fatalErr) {
+			fmt.Fprintf(os.Stderr, "Error: %v\n", err)
+			os.Exit(1)
+		}
+		// Non-fatal update failures: MaybeUpdateAndReexec already logged to stderr and returned nil
+		// If we get here with a non-fatal error, it's a reexec error or similar - log and continue
 		fmt.Fprintf(os.Stderr, "Error: %v\n", err)
-		os.Exit(1)
 	}
 	if err := rootCmd.Execute(); err != nil {
 		if !errors.Is(err, jsonout.ErrAlreadyWritten) {
--- a/cmd/liza/main.go
+++ b/cmd/liza/main.go
@@ -236,6 +236,8 @@ func init() {
 	// This allows update checks to run before the main command, while still showing these flags
 	// in --help output. The manual parsing in internal/updater implements stop-at-double-dash
 	// and last-flag-wins semantics to match pflag/Cobra behavior for these specific flags.
+	// Update channel validation (stable/main) is performed during this pre-parsing phase,
+	// and invalid values cause a fatal error before Cobra command execution.
 	rootCmd.PersistentFlags().Bool("check-update", false, "check for a Liza update before running")
 	rootCmd.PersistentFlags().String("update-channel", "stable", "update check channel: stable or main")
 }
--- a/cmd/liza/main.go
+++ b/cmd/liza/main.go
@@ -1,6 +1,7 @@
 package main
 
 import (
+	"context"
 	"errors"
 	"fmt"
 	"os"
@@ -9,10 +10,12 @@ import (
 
 	lizaerrors "github.com/liza-mas/liza/internal/errors"
 	"github.com/liza-mas/liza/internal/identity"
+	"github.com/liza-mas/liza/internal/interactive"
 	"github.com/liza-mas/liza/internal/jsonout"
 	"github.com/liza-mas/liza/internal/ops"
 	"github.com/liza-mas/liza/internal/paths"
 	"github.com/liza-mas/liza/internal/pipeline"
+	"github.com/liza-mas/liza/internal/updater"
 	"github.com/spf13/cobra"
 )
 
@@ -228,6 +231,8 @@ func init() {
 	// Global flags
 	rootCmd.PersistentFlags().BoolP("verbose", "v", false, "verbose output")
 	rootCmd.PersistentFlags().StringP("project-root", "C", "", "Liza project root for state commands")
+	rootCmd.PersistentFlags().Bool("check-update", false, "check for a Liza update before running")
+	rootCmd.PersistentFlags().String("update-channel", "stable", "update check channel: stable or main")
 }
 
 // addAgentIDFlag registers --agent-id on a specific command.
@@ -253,6 +258,14 @@ func addChangedByFlag(cmd *cobra.Command) {
 
 func main() {
 	if err := checkSupportedPlatform(runtime.GOOS); err != nil {
+		fmt.Fprintf(os.Stderr, "Error: %v\n", err)
+		os.Exit(1)
+	}
+	if err := updater.MaybeUpdateAndReexec(context.Background(), updater.Config{
+		CurrentVersion: Version,
+		CurrentCommit:  GitCommit,
+		IsInteractive:  interactive.IsInteractive,
+	}); err != nil {
 		fmt.Fprintf(os.Stderr, "Error: %v\n", err)
 		os.Exit(1)
 	}
--- a/internal/updater/updater.go
+++ b/internal/updater/updater.go
@@ -145,7 +145,17 @@ func MaybeUpdateAndReexec(ctx context.Context, cfg Config) error {
 	installCtx, cancel := context.WithTimeout(ctx, cfg.InstallTimeout)
 	defer cancel()
 
+	rollback, err := prepareInstallRollback(target)
+	if err != nil {
+		fmt.Fprintf(cfg.Stderr, "liza: prepare update rollback failed: %v\n", err)
+		return nil
+	}
+	defer rollback.cleanup()
+
 	if err := cfg.Install(installCtx, next, target, cfg.Stderr, cfg.Stderr); err != nil {
+		if restoreErr := rollback.restore(target); restoreErr != nil {
+			fmt.Fprintf(cfg.Stderr, "liza: restore previous install failed: %v\n", restoreErr)
+		}
 		fmt.Fprintf(cfg.Stderr, "liza: update install failed: %v\n", err)
 		return nil
 	}
@@ -156,6 +166,9 @@ func MaybeUpdateAndReexec(ctx context.Context, cfg Config) error {
 		defer cancel()
 		if err := cfg.VerifyInstall(verifyCtx, target, cfg.Stderr); err != nil {
 			fmt.Fprintf(cfg.Stderr, "liza: post-install verification failed: %v\n", err)
+			if restoreErr := rollback.restore(target); restoreErr != nil {
+				fmt.Fprintf(cfg.Stderr, "liza: restore previous install failed: %v\n", restoreErr)
+			}
 			fmt.Fprintf(cfg.Stderr, "liza: falling through to original command without reexec\n")
 			return nil
 		}
@@ -472,6 +485,74 @@ func replaceBinary(target string, r io.Reader, mode os.FileMode) error {
 	}
 	cleanup = false
 	return nil
+}
+
+type installRollback struct {
+	backupPath string
+}
+
+func prepareInstallRollback(target string) (installRollback, error) {
+	info, err := os.Stat(target)
+	if err != nil {
+		if os.IsNotExist(err) {
+			return installRollback{}, nil
+		}
+		return installRollback{}, fmt.Errorf("stat current binary: %w", err)
+	}
+	if info.IsDir() {
+		return installRollback{}, fmt.Errorf("current binary target is a directory: %s", target)
+	}
+
+	src, err := os.Open(target)
+	if err != nil {
+		return installRollback{}, fmt.Errorf("open current binary: %w", err)
+	}
+	defer src.Close()
+
+	tmp, err := os.CreateTemp(filepath.Dir(target), ".liza-update-backup-*")
+	if err != nil {
+		return installRollback{}, fmt.Errorf("create backup binary: %w", err)
+	}
+	backupPath := tmp.Name()
+	cleanup := true
+	defer func() {
+		if cleanup {
+			_ = os.Remove(backupPath)
+		}
+	}()
+
+	if _, err := io.Copy(tmp, src); err != nil {
+		_ = tmp.Close()
+		return installRollback{}, fmt.Errorf("copy current binary to backup: %w", err)
+	}
+	if err := tmp.Close(); err != nil {
+		return installRollback{}, fmt.Errorf("close backup binary: %w", err)
+	}
+	if err := os.Chmod(backupPath, info.Mode().Perm()); err != nil {
+		return installRollback{}, fmt.Errorf("chmod backup binary: %w", err)
+	}
+
+	cleanup = false
+	return installRollback{backupPath: backupPath}, nil
+}
+
+func (r *installRollback) restore(target string) error {
+	if r.backupPath == "" {
+		return nil
+	}
+	if err := os.Rename(r.backupPath, target); err != nil {
+		return fmt.Errorf("restore %s from backup: %w", target, err)
+	}
+	r.backupPath = ""
+	return nil
+}
+
+func (r *installRollback) cleanup() {
+	if r.backupPath == "" {
+		return
+	}
+	_ = os.Remove(r.backupPath)
+	r.backupPath = ""
 }
 
 func installFromSource(ctx context.Context, commit, target string, stderr io.Writer) error {
--- a/internal/updater/updater.go
+++ b/internal/updater/updater.go
@@ -87,6 +87,10 @@ type moduleInfo struct {
 	} `json:"Origin"`
 }
 
+type releaseInfo struct {
+	TagName string `json:"tag_name"`
+}
+
 type candidate struct {
 	Channel string
 	Current string
@@ -207,11 +211,23 @@ func lookupCandidate(ctx context.Context, cfg Config) (candidate, error) {
 }
 
 func LatestModuleVersion(ctx context.Context) (string, error) {
-	out, err := runOutput(ctx, "go", "list", "-m", "-json", modulePath+"@latest")
+	req, err := http.NewRequestWithContext(ctx, http.MethodGet, "https://api.github.com/repos/liza-mas/liza/releases/latest", nil)
 	if err != nil {
 		return "", err
 	}
-	return parseLatestVersion(out)
+	resp, err := http.DefaultClient.Do(req)
+	if err != nil {
+		return "", fmt.Errorf("lookup latest release: %w", err)
+	}
+	defer resp.Body.Close()
+	if resp.StatusCode != http.StatusOK {
+		return "", fmt.Errorf("lookup latest release: HTTP %s", resp.Status)
+	}
+	data, err := io.ReadAll(resp.Body)
+	if err != nil {
+		return "", fmt.Errorf("read latest release response: %w", err)
+	}
+	return parseLatestReleaseVersion(data)
 }
 
 func MainCommit(ctx context.Context) (string, error) {
@@ -514,6 +530,18 @@ func parseLatestVersion(data []byte) (string, error) {
 	version := normalizeVersion(info.Version)
 	if !semver.IsValid(version) {
 		return "", fmt.Errorf("latest module version is not semantic: %q", info.Version)
+	}
+	return version, nil
+}
+
+func parseLatestReleaseVersion(data []byte) (string, error) {
+	var info releaseInfo
+	if err := json.Unmarshal(data, &info); err != nil {
+		return "", fmt.Errorf("parse latest release response: %w", err)
+	}
+	version := normalizeVersion(info.TagName)
+	if !semver.IsValid(version) {
+		return "", fmt.Errorf("latest release tag is not semantic: %q", info.TagName)
 	}
 	return version, nil
 }
--- a/internal/updater/updater.go
+++ b/internal/updater/updater.go
@@ -1,0 +1,617 @@
+package updater
+
+import (
+	"archive/tar"
+	"bufio"
+	"bytes"
+	"compress/gzip"
+	"context"
+	"encoding/json"
+	"fmt"
+	"io"
+	"net/http"
+	"os"
+	"os/exec"
+	"path/filepath"
+	"runtime"
+	"strings"
+	"syscall"
+	"time"
+
+	"github.com/liza-mas/liza/internal/paths"
+	"golang.org/x/mod/semver"
+)
+
+const (
+	modulePath  = "github.com/liza-mas/liza"
+	commandPath = modulePath + "/cmd/liza"
+
+	channelStable = "stable"
+	channelMain   = "main"
+
+	channelEnvName = "LIZA_UPDATE_CHANNEL"
+	checkEnvName   = "LIZA_CHECK_UPDATE"
+	skipEnvName    = "LIZA_SKIP_AUTO_UPDATE"
+	updatePrefs    = "update.json"
+)
+
+type Config struct {
+	CurrentVersion string
+	CurrentCommit  string
+	Args           []string
+	Env            []string
+	Stdin          io.Reader
+	Stdout         io.Writer
+	Stderr         io.Writer
+	IsInteractive  func() bool
+	CheckUpdate    bool
+	Channel        string
+	LookupLatest   func(context.Context) (string, error)
+	LookupMain     func(context.Context) (string, error)
+	Install        func(context.Context, candidate, string, io.Writer, io.Writer) error
+	InstallTarget  func() (string, error)
+	CheckDisabled  func() bool
+	DisableCheck   func() error
+	Reexec         func(string, []string, []string) error
+}
+
+type moduleInfo struct {
+	Version string `json:"Version"`
+	Origin  struct {
+		Hash string `json:"Hash"`
+	} `json:"Origin"`
+}
+
+type candidate struct {
+	Channel string
+	Current string
+	Latest  string
+	Ref     string
+}
+
+func MaybeUpdateAndReexec(ctx context.Context, cfg Config) error {
+	cfg = withDefaults(cfg)
+	if shouldSkip(cfg) {
+		return nil
+	}
+
+	lookupCtx, cancel := context.WithTimeout(ctx, 30*time.Second)
+	defer cancel()
+
+	next, err := lookupCandidate(lookupCtx, cfg)
+	if err != nil {
+		fmt.Fprintf(cfg.Stderr, "liza: update check failed: %v\n", err)
+		return nil
+	}
+	if next.Ref == "" {
+		return nil
+	}
+
+	input := bufio.NewReader(cfg.Stdin)
+	if !confirmUpdate(input, cfg.Stdout, next) {
+		proposeDisableCheck(input, cfg.Stdout, cfg.DisableCheck)
+		return nil
+	}
+
+	target, err := cfg.InstallTarget()
+	if err != nil {
+		return fmt.Errorf("find liza install target: %w", err)
+	}
+	fmt.Fprintln(cfg.Stdout, installPlan(next, target))
+	if err := cfg.Install(ctx, next, target, cfg.Stdout, cfg.Stderr); err != nil {
+		return fmt.Errorf("update install: %w", err)
+	}
+
+	args := reexecArgs(cfg.Args)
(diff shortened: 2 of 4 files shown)


tangi-vass (changes requested):
Review: Standard — Request Changes

Blockers: 1
Concerns: 3
Suggestions: 0

[blocker] internal/updater/updater.go:203 — Stable update replaces the executable from an unverified downloaded archive

Why it matters: this is a self-update path that writes over the running CLI binary. If the archive is corrupted, swapped, or served from an unintended `LIZA_RELEASE_BASE_URL`, the updater can install untrusted code. Releases already publish `checksums.txt` via `.goreleaser.yaml`, so the updater has a verification source available.

Suggestion: download and verify the archive checksum before extraction/replacement, and preferably move toward signed artifacts/provenance for this path.

[concern] internal/updater/updater.go:101 / cmd/liza/main.go:264 — Install runs with an unbounded `context.Background()`

Why it matters: lookup has a 30s timeout, but once the user accepts, HTTP download, `git clone`, checkout, and `make install` can hang indefinitely and block the CLI.

Suggestion: use a bounded install context and add cancellation/timeout tests.

[concern] internal/updater/updater.go:82 / internal/updater/updater.go:496 — Invalid explicit update flags only warn and continue

Why it matters: `liza --check-update --update-channel=bogus version` is explicit user input, but `MaybeUpdateAndReexec` prints an updater warning and then runs the original command. That makes invalid CLI usage look successful.

Suggestion: when the check was explicitly requested by flag/config, return a CLI error for invalid channel. Also fix the manual parser to honor `--` and last-flag-wins semantics.

[concern] internal/updater/updater.go:377 / internal/interactive/tty.go:8 — Update prompts use stdout while interactivity only checks stdin

Why it matters: with `LIZA_CHECK_UPDATE=1`, a terminal stdin plus redirected stdout or JSON command output can receive update prompt text on stdout before the command output, corrupting machine-readable output.

Suggestion: require stdout to be a TTY for prompting, or write update prompts/status to stderr and skip update prompting for structured-output flows.

Context checked: PR description, Copilot review + inline comments, issue comments, and attachment/url scan. No PR attachments found; only the Copilot docs URL appeared.

Validation run locally: `go test ./internal/updater ./cmd/liza` passed: 240 tests in 2 packages. I did not run full `make test` because it invokes `sync-embedded`, which would mutate this already-dirty review workspace.

internal/updater/updater.go
```diff
@@ -0,0 +1,617 @@
...
+	}
+	if err := runStreaming(ctx, stdout, stderr, "git", "-C", repoDir, "checkout", commit); err != nil {
+		return fmt.Errorf("checkout liza source %s: %w", commit, err)
+	}
+	installDir := filepath.Dir(target)
+	if err := runStreaming(ctx, stdout, stderr, "make", "-C", repoDir, "install", "INSTALL_DIR="+installDir); err != nil {
+		return fmt.Errorf("build liza source: %w", err)
+	}
+	return nil
+}
```
tangi-vass, line 588: **[blocker]** `installFromSource` shallow clone cannot reach arbitrary commit

`git clone --depth 1 --branch main` fetches only the tip of `main`. The subsequent `git checkout <commit>` will fail for any commit that isn't exactly HEAD of main, because that commit isn't in the shallow history. This means the main-channel update path is broken for any non-trivial case where the user is even one commit behind `main` tip while a newer commit exists.

The manual verification described in the PR description likely passed because the "fake-cloned repo" already contained the target commit.

Suggestion: Either `git clone --depth 1` then `git fetch --depth 1 origin <commit>` + `git checkout FETCH_HEAD`, or skip depth limiting entirely for correctness.

internal/updater/updater.go
```diff
@@ -0,0 +1,617 @@
...
+	resp, err := http.DefaultClient.Do(req)
+	if err != nil {
+		return fmt.Errorf("download release archive: %w", err)
+	}
+	defer resp.Body.Close()
+	if resp.StatusCode != http.StatusOK {
+		return fmt.Errorf("download release archive: HTTP %s", resp.Status)
+	}
+	return installBinaryFromTarGz(resp.Body, target)
+}
```
tangi-vass, line 339: **[blocker]** No integrity verification on downloaded binary

The stable-channel updater downloads a tarball over HTTPS and replaces the running binary with no checksum or signature verification. `LIZA_RELEASE_BASE_URL` allows redirecting downloads to any server. A compromised CDN, MITM on TLS, or a user-set env var pointing to a malicious server would result in arbitrary code execution on the next `syscall.Exec`.

At minimum, the release archive should be verified against a published checksum (SHA256 in a `.sha256` sidecar or from the GitHub release API). This can be a follow-up if tracked as tech debt with a concrete payback trigger ("before first public release"), but it needs to be explicitly acknowledged.

internal/updater/updater.go
```diff
@@ -0,0 +1,617 @@
...
+		return true
+	}
+	if !checkUpdateEnvEnabled(cfg) {
+		return true
+	}
+	if !cfg.IsInteractive() {
+		return true
+	}
+	return false
+}
```
tangi-vass, line 712: **[concern]** `shouldSkip` precedence is confusing and partially wrong

Two parallel flag-reading mechanisms exist: cobra flags registered in `init()` (visible in `--help`) and manual `os.Args` parsing here. In `main.go`, the cobra flags are registered but never read from cobra — `Config.CheckUpdate` is always `false` in production. The `cfg.CheckUpdate` branch (line 395-397) is unreachable in production.

This works by accident because `checkUpdateFlag` re-parses `os.Args` directly. But the cobra flags are visible in `--help` and their registered defaults never flow to the updater.

Suggestion: Either remove the cobra flag registration and document that these are raw-args-only flags (since the updater runs before `rootCmd.Execute()`), or read the cobra flags in a `PersistentPreRun` and pass them to the updater. The current hybrid will confuse contributors.

internal/updater/updater.go
```diff
@@ -0,0 +1,617 @@
...
+		}
+	}
+	if cfg.Channel != "" {
+		return strings.ToLower(strings.TrimSpace(cfg.Channel))
+	}
+	if envChannel := envValue(cfg.Env, channelEnvName); envChannel != "" {
+		return strings.ToLower(strings.TrimSpace(envChannel))
+	}
+	return channelStable
+}
```
tangi-vass, line 856: **[concern]** Neither `checkUpdateFlag` nor `updateChannel` stop at `--`

Args after `--` are pass-through to subcommands and should not be interpreted as updater flags. `liza agent orchestrator -- --check-update` would incorrectly trigger update checks. Same issue affects `--update-channel`.

Copilot flagged this too — it's a real issue worth fixing since `liza agent` passes arguments through.

cmd/liza/main.go
```diff
@@ -256,6 +261,14 @@ func main() {
...
 		os.Exit(1)
 	}
+	if err := updater.MaybeUpdateAndReexec(context.Background(), updater.Config{
+		CurrentVersion: Version,
+		CurrentCommit:  GitCommit,
+		IsInteractive:  interactive.IsInteractive,
+	}); err != nil {
+		fmt.Fprintf(os.Stderr, "Error: %v\n", err)
+		os.Exit(1)
+	}
```
tangi-vass, line 285: **[concern]** Update failure exits the entire CLI

`MaybeUpdateAndReexec` returning an error causes `os.Exit(1)`. This means if `InstallTarget` fails, or the install itself fails mid-way, the user cannot run any liza command at all until the condition is resolved — including `liza status`, `liza stop`, and other operational commands.

For an opt-in convenience feature, a failed update should not prevent the original command from executing.

Suggestion: Log the error to stderr and fall through to `rootCmd.Execute()` instead of exiting.

internal/updater/updater.go
```diff
@@ -0,0 +1,617 @@
...
+	"syscall"
+	"time"
+
+	"github.com/liza-mas/liza/internal/paths"
+	"golang.org/x/mod/semver"
+)
+
+const (
+	modulePath  = "github.com/liza-mas/liza"
+	commandPath = modulePath + "/cmd/liza"
```
tangi-vass: **[concern]** `commandPath` is declared but never used — dead code.

internal/updater/updater.go
```diff
@@ -0,0 +1,617 @@
...
+	}
+	fmt.Fprintln(cfg.Stdout, installPlan(next, target))
+	if err := cfg.Install(ctx, next, target, cfg.Stdout, cfg.Stderr); err != nil {
+		return fmt.Errorf("update install: %w", err)
+	}
+
+	args := reexecArgs(cfg.Args)
+	env := setEnv(cfg.Env, skipEnvName, "1")
+	return cfg.Reexec(target, args, env)
+}
```
tangi-vass, line 180: **[suggestion]** Consider verifying the new binary before reexec

`MaybeUpdateAndReexec` does check → prompt → install → reexec. The install+reexec path is irreversible (binary replaced, process replaced via `syscall.Exec`). If the install succeeds but the binary is corrupt, the user is stuck.

Consider running `<target> version` to verify the new binary can execute before calling `Reexec`.

internal/updater/updater.go
```diff
@@ -0,0 +1,617 @@
...
+	if resp.StatusCode != http.StatusOK {
+		return fmt.Errorf("download release archive: HTTP %s", resp.Status)
+	}
+	return installBinaryFromTarGz(resp.Body, target)
+}
+
+func releaseArchiveURL(version, goos, goarch string) string {
+	versionBare := strings.TrimPrefix(version, "v")
+	archive := fmt.Sprintf("liza-%s-%s-%s.tar.gz", versionBare, goos, goarch)
+	baseURL := strings.TrimRight(os.Getenv("LIZA_RELEASE_BASE_URL"), "/")
```
tangi-vass: **[suggestion]** `releaseArchiveURL` reads `os.Getenv("LIZA_RELEASE_BASE_URL")` directly, while other env access goes through the `cfg.Env` slice via `envValue`. This makes the base URL untestable through Config and is inconsistent with the rest of the file.

tangi-vass (changes requested):
## Review: Deep — Request Changes

**Blockers: 2 | Concerns: 4 | Suggestions: 3**

The architecture is sound and the test coverage is good for the skip/enable matrix. The Config dependency injection with `withDefaults` is well-structured, and the `replaceBinary` write-to-temp-then-rename pattern is the correct approach for atomic binary replacement. The re-exec loop prevention via `LIZA_SKIP_AUTO_UPDATE` is clean.

Two issues need resolution before merge: the shallow-clone bug makes the main channel non-functional for non-tip commits, and the binary download lacks integrity verification. The flag-passing mechanism and error handling also need cleanup.

**Blast Radius:** High — self-updating binary replacement, CLI startup path, `syscall.Exec`
**Confidence:** High — thorough review of all 4 files
**Next step:** Fix the shallow-clone bug and decide on the integrity verification timeline, then re-review focused on those fixes.

### Copilot comments assessment

| Copilot comment | Assessment |
|---|---|
| `checkUpdateFlag` doesn't stop at `--` | Valid — covered in concern below |
| `updateChannel` doesn't stop at `--` | Valid — same concern |
| Lookup errors silently ignored | Partially valid — for network errors log+continue is reasonable; for invalid channel, the error is swallowed |
| Install phase has no timeout | Valid but low-severity — user is interactive and can Ctrl-C |
| No integrity verification | Valid — elevated to blocker |

tangi-vass (commented):
## Review: Deep (re-review) — Comment

**Blockers: 0 | Concerns: 1 | Suggestions: 3**

Re-review scoped to the fix commit `ddb1d454` against the two prior reviews (Standard + Deep). All prior blockers and most concerns are resolved. Good work on the fix — it's thorough and well-tested.

### Prior findings disposition

| Prior finding | Status | Notes |
|---|---|---|
| **[blocker]** Shallow clone cannot reach non-tip commits | **Fixed** | `git fetch --depth 1 origin <commit>` + `checkout --detach FETCH_HEAD`. `TestInstallFromSourceFetchesExactCommit` proves the pattern. |
| **[blocker]** No integrity verification on binary | **Fixed** | SHA256 checksum verification against `checksums.txt`. Tests cover valid, mismatch, and missing checksum. |
| **[concern]** Update failure exits entire CLI | **Fixed** | `FatalError` type distinguishes invalid config (exit) from install/lookup failures (log to stderr, fall through). `main.go:274-282` handles the split correctly. |
| **[concern]** Flag parsers don't stop at `--` | **Fixed** | Both `checkUpdateFlag` and `updateChannel` stop at `--`. Tests cover this + last-flag-wins. |
| **[concern]** Hybrid cobra+manual parsing confusing | **Fixed** | 5-line comment in `init()` explains the design and why. |
| **[concern]** Update prompts to stdout | **Fixed** | Prompts now go to `cfg.Stderr`. Tests updated. |
| **[concern]** Update failure exits CLI (Copilot) | **Fixed** | Same as above — non-fatal failures return nil after logging. |
| **[concern]** Dead `commandPath` | **Fixed** | Removed. |
| **[suggestion]** `releaseArchiveURL` uses `os.Getenv` directly | **Partially fixed** | Uses `releaseBaseEnvName` constant now, but still `os.Getenv` not `cfg.Env`. See suggestion below. |
| **[suggestion]** Verify new binary before reexec | **Not addressed** | Acceptable deferral — the checksum verification reduces the risk significantly. |

### New findings on the fix commit

**[concern]** `updater.go:251-273` — Archive and checksums downloaded from the same origin

Both the archive and its `checksums.txt` are fetched from the same `LIZA_RELEASE_BASE_URL`. If an attacker controls that URL (via env var or DNS), they serve both the malicious archive and a matching checksum file, bypassing verification entirely.

This is a known limitation of checksum-only verification vs. signature-based verification (e.g., cosign/Sigstore). Not a blocker — the PR's scope was "at minimum, verify against published checksums" and it delivers that. But it should be tracked as tech debt with payback trigger "before first public release" if not already.

Why it matters: The checksum verification provides integrity (detect corruption/MITM on TLS) but not authenticity (prove the release came from the project). For a self-updating binary, authenticity is the higher-value property.

**[suggestion]** `updater.go:281,289` — `releaseArchiveURL` and `checksumURL` duplicate the base URL resolution

Both functions independently read `os.Getenv(releaseBaseEnvName)` and apply the same fallback. Extract a `releaseBaseURL()` helper.

**[suggestion]** `updater.go:395-425` / `updater_test.go:395-425` — Tests use `os.Getenv`/`os.Setenv` pattern instead of `t.Setenv`

`TestReleaseBaseURLControlsBothArchiveAndChecksum`, `TestInstallReleaseBinaryWithChecksum`, `TestInstallReleaseBinaryChecksumMismatch`, and `TestInstallReleaseBinaryMissingChecksum` all manually save/restore env vars. `t.Setenv` (Go 1.17+) is cleaner and automatically restores on test cleanup. Minor — the current approach works, but `t.Setenv` is the Go idiom used elsewhere in the codebase.

**[suggestion]** `updater.go:281,289` — `releaseArchiveURL` and `checksumURL` still read `os.Getenv` directly

Prior suggestion from Round 1. All other env access goes through `cfg.Env` via `envValue`, making it testable through Config injection. The base URL bypasses this, forcing tests to use `os.Setenv`. Could be addressed by passing `cfg.Env` into these functions, or by adding a `ReleaseBaseURL` field to Config.

### Appraisals

**[appraisal]** The fix commit is well-structured — it addresses each finding with a targeted change and adds test coverage for each fix. The `TestInstallFromSourceFetchesExactCommit` test is particularly good: it creates a real multi-commit repo, proves the shallow clone fails for non-tip commits, and then proves the fetch+FETCH_HEAD pattern succeeds. That's the kind of test that prevents regression.

**[appraisal]** The `FatalError` type is a clean solution for the "invalid config should exit, install failure should continue" distinction. The error handling in `MaybeUpdateAndReexec` is now correct: fatal errors propagate, everything else logs and returns nil, and only reexec errors reach `main.go` as non-fatal (which is the right behavior since the binary was already replaced).

---

**Overall:** All blockers resolved. The fix commit is thorough and well-tested. The remaining concern (same-origin checksums) is a known limitation worth tracking but not blocking. Suggestions are minor cleanup.

**Blast Radius:** High — unchanged (self-updating binary replacement, CLI startup path, `syscall.Exec`)
**Confidence:** High — reviewed all changes in the fix commit against prior findings, verified each fix addresses the root cause
**Verdict:** Approve with the note to track signature-based verification as tech debt.

tangi-vass (changes requested):
Review: Deep focused re-review - Request Changes

Blockers: 2
Concerns: 1
Suggestions: 0

[blocker] internal/updater/updater_test.go:1072 - PR head does not compile

Why it matters: `go test ./internal/updater` fails at build time:
`installCtx` declared and not used, and `now` declared and not used at line 1166. CI is already failing on Ubuntu.

Suggestion: remove the unused vars or assert the values they were meant to prove.

[blocker] internal/updater/updater_test.go:772 - source-update regression test does not exercise the real shallow-fetch fix

Why it matters: the mock checks `args[3] == "fetch"`, but `installFromSource` calls `git -C <repo> fetch ...`, so `"fetch"` is at `args[2]`. The shallow-fetch failure is never simulated. Once the compile errors are fixed, this test will either fail or pass without proving the production path.

Suggestion: assert the actual command shape, or make the source repo URL injectable and test `installFromSource` against a `file://` local repo where `--depth 1` is honored.

[concern] internal/updater/updater.go:288 / internal/updater/updater_test.go:491 - checksum verification is not test-injectable end to end

Why it matters: `installReleaseBinary` downloads the archive from the injected server, but always fetches `checksums.txt` from canonical GitHub. The new httptest checksum endpoint is unreachable, so the checksum install tests will not validate the stable install path after the compile errors are fixed.

Suggestion: inject checksum lookup separately, or keep canonical production behavior but add a test-only path that exercises archive + checksum verification together.

Prior review status: the production code now addresses the main themes: stderr prompts, non-fatal install failures, bounded install context, checksum verification, and exact commit fetch. The remaining issue is that the verification layer is broken enough that the PR is not mergeable yet.

Context checked: PR description, reviews, inline comments, issue comments, URL/attachment scan. No PR attachments found; only the Copilot docs URL appeared.

Validation:
- `go test ./internal/updater` failed with the compile errors above.
- `make sync-embedded && go test ./cmd/liza` passed in a temp PR checkout.

tangi-vass (changes requested):
Review: Deep focused re-review - Request Changes

Blockers: 3
Concerns: 0
Suggestions: 0

[blocker] internal/updater/updater_test.go:1072 / internal/updater/updater_test.go:1166 - PR head does not compile

Why it matters: `go test ./internal/updater` fails at build time on current head `062f7bc`:
`installCtx` is declared and not used, and `now` is declared and not used. This also explains the current Ubuntu CI failure.

Suggestion: remove the unused locals or assert the values they were intended to prove.

[blocker] internal/updater/updater_test.go:209 / internal/updater/updater_test.go:267 / internal/updater/updater_test.go:321 / internal/updater/updater_test.go:1071 / internal/updater/updater_test.go:1126 - reexec tests are stale after `VerifyInstall` became the default

Why it matters: after a temp-only patch to bypass the two compile errors, `go test ./internal/updater` still fails these tests because they fake a successful `Install` but leave `VerifyInstall` at the production default. `MaybeUpdateAndReexec` now verifies the target binary before reexec; the fake targets do not exist or are not runnable, so the function logs verification failure, returns nil, and never reaches the `Reexec` sentinel.

Suggestion: for tests that are only asserting reexec wiring, pass `VerifyInstall: func(context.Context, string, io.Writer) error { return nil }`. For tests that should cover verification, create a temp executable and assert both success and failure behavior explicitly.

[blocker] internal/updater/updater_test.go:772 / internal/updater/updater_test.go:1173 - source-fetch tests still do not reliably prove exact-commit source updates

Why it matters: after the compile-only bypass, `TestInstallFromSourceShallowFetchFallback` fails because its mock checks `args[3] == "fetch"`, but production calls `git -C <repo> fetch ...`, so `fetch` is at `args[2]`. The shallow-fetch failure is never simulated. `TestInstallFromSourceFetchesExactCommit` also fails in this environment at `git clone --branch main` because the test repo is initialized without forcing a `main` branch; even after that, cloning from a local path can ignore `--depth`, so it may not prove the shallow-history failure mode.

Suggestion: make the source repo URL injectable and exercise `installFromSource` against a `file://` repo initialized with `git init -b main`, or tighten the mock to assert the exact command sequence and argument positions.

Prior findings status: the production code now appears to address the earlier updater concerns: stderr prompts, non-fatal install failures, bounded install context, checksum verification, exact commit fetch, and post-install verification. The remaining problem is that the updater verification layer is not mergeable because the tests do not compile and still fail once the compile errors are bypassed.

Context checked: PR description, prior reviews, inline comments, issue comments, URL/attachment scan. No PR issue comments or attachments found; only the Copilot docs URL appeared.

Validation:
- `go test ./internal/updater` failed on current head with the compile errors above.
- After a temp-only patch in `/tmp` to bypass only those two unused-variable errors, `go test ./internal/updater` failed with 62 passed / 7 failed, including the stale `VerifyInstall` expectations and source-fetch tests above.
- `make sync-embedded && go test ./cmd/liza` passed in the temp PR checkout.

tangi-vass (commented):
Review: Deep focused re-review - Comment

Blockers: 0
Concerns: 1
Suggestions: 0

Prior findings status: the current head `c4ef86a9` appears to resolve the prior merge blockers. The updater package now compiles, the stale `VerifyInstall` expectations in the reexec tests are fixed, the source-fetch fallback test exercises the real `git -C <repo> fetch --depth 1 ...` command shape, the exact-commit source test forces `main`, and the checksum tests are platform-independent.

[concern] internal/updater/updater.go:153 - Verification failure does not roll back the already-replaced binary

Why it matters: `cfg.Install` has already replaced the target by the time `VerifyInstall` runs. If verification fails, `MaybeUpdateAndReexec` only falls through to the original command. The current process keeps working, but the on-disk `liza` can remain broken for the next invocation. For a self-update path, "verification failed" should mean the installed CLI remains usable or the old binary is restored.

Suggestion: make installs two-phase: install into a temp target, run `VerifyInstall` against that temp binary, then atomically rename over the real target only after verification succeeds. For source installs, use a temporary `INSTALL_DIR` and move the resulting `liza` into place. If two-phase install is too large for this PR, keep a backup of the old binary and restore it on verification failure.

Context checked: PR description, all review/inline/issue comments, and URL/attachment scan. No PR issue comments or attachments found; only the Copilot docs URL appeared.

Validation:
- `go test ./internal/updater ./cmd/liza` passed locally: 292 tests in 2 packages.
- PR CI checks are green on Ubuntu and macOS.

Overall: The prior test/build blockers appear resolved. The remaining functional risk is rollback semantics around failed post-install verification.
Blast Radius: High - self-updating binary replacement, CLI startup path, `syscall.Exec`
Confidence: High - focused re-review of latest fix commits and current updater flow.
Next step: decide whether to handle rollback before merge or explicitly track it as tech debt with a payback trigger before enabling/publicizing stable auto-update.

tangi-vass (approved):
Review: Deep focused re-review - Approve

Blockers: None
Concerns: None
Suggestions: 0

Latest commit checked: `c7bcc7c8` (`fix(updater): restore binary on verification failure`).

The prior rollback concern is resolved. `MaybeUpdateAndReexec` now prepares a backup before install, restores it on install failure, and restores it on post-install verification failure before falling through to the original command. The new test `TestMaybeUpdateAndReexecRestoresTargetWhenVerificationFails` exercises the failure mode directly: it replaces the target with a broken body, forces verification failure, asserts reexec is skipped, and asserts the original target body is restored.

Validation:
- Local: `go test ./internal/updater ./cmd/liza` passed: 293 tests in 2 packages.
- PR CI: Ubuntu and macOS checks are green on head `c7bcc7c8`.

Overall: the self-update path now has checksum verification, exact source commit fetching, bounded install/verify contexts, non-fatal operational failures, post-install verification, and rollback on failed verification. I do not see remaining blockers or concerns for this PR.

Blast Radius: High - self-updating binary replacement, CLI startup path, `syscall.Exec`
Confidence: High - focused re-review of the latest commit plus scoped validation.
Next step: merge when ready.

tangi-vass:
## Review: Deep — Comment (Approve)

**Blockers: 0 | Concerns: 1 | Suggestions: 2**

Reviewing in Deep mode because this is a self-updating binary replacement path — high blast radius touching CLI startup, `syscall.Exec`, and on-disk binary integrity.

### Architecture assessment

The updater is well-architected. The `Config` struct provides clean dependency injection for all external effects (lookup, install, verify, reexec), making the core flow testable without network or filesystem side effects. The `withDefaults` pattern correctly separates production wiring from testable logic. The `FatalError` type cleanly splits "invalid user input → exit" from "transient failure → log and continue", which is the right UX for a pre-command interceptor.

The flow — check → prompt → backup → install → verify → reexec (or rollback) — covers the important failure modes in the right order. The re-exec loop prevention via `LIZA_SKIP_AUTO_UPDATE` is simple and correct.

### Prior review history (7 rounds) — disposition

| Finding | Status |
|---|---|
| **[blocker]** No integrity verification on binary download | **Fixed** — SHA256 checksum verification against `checksums.txt` with tests for valid/mismatch/missing |
| **[blocker]** Shallow clone cannot reach non-tip commits | **Fixed** — `git fetch --depth 1 origin <commit>` + `checkout --detach FETCH_HEAD` with deep-fetch fallback |
| **[blocker]** PR head does not compile (unused vars) | **Fixed** — unused locals removed |
| **[blocker]** Reexec tests stale after `VerifyInstall` default | **Fixed** — tests inject no-op `VerifyInstall` where appropriate |
| **[blocker]** Source-fetch mock arg position off-by-one | **Fixed** — test exercises real command shape |
| **[concern]** Update failure exits entire CLI | **Fixed** — `FatalError` for config errors, nil return for transient failures |
| **[concern]** Flag parsers don't stop at `--` | **Fixed** — both stop at `--`, tests cover this + last-flag-wins |
| **[concern]** Update prompts to stdout | **Fixed** — all prompts/status go to `cfg.Stderr` |
| **[concern]** Same-origin checksums | **Accepted** — known limitation, tracked |
| **[concern]** Verification failure doesn't roll back binary | **Fixed** — `prepareInstallRollback` backs up the old binary before install, restores on install or verification failure |
| **[suggestion]** `os.Getenv` for release base URL | **Fixed** — `ReleaseBaseURL` is now a `Config` field, test-injectable |
| **[suggestion]** Verify new binary before reexec | **Fixed** — `VerifyInstall` runs `<target> version` post-install |
| **[suggestion]** `t.Setenv` instead of manual save/restore | Not addressed — minor, non-blocking |

### New findings

**[concern]** `updater.go:155-161` / `updater.go:166-174` — Install failure rollback logs but still returns nil; partial writes to the binary may leave it in an inconsistent state if `cfg.Install` fails mid-write *and* `rollback.restore` also fails (e.g., same filesystem permission issue that caused the original failure)

Why it matters: If the install overwrites the target binary partway through and then fails, and then restore also fails (e.g., directory permissions changed, disk full), the on-disk `liza` is corrupt and the user sees two error messages but the CLI proceeds with the original command, giving the impression everything is fine. The next `liza` invocation will fail with an opaque exec error.

This is mitigated by: (1) `replaceBinary` uses temp-file-then-rename, so partial writes don't corrupt the target — only a successful rename replaces it; (2) the rollback covers the rename-succeeded-but-verification-failed case. The remaining gap is narrow: `os.Rename` succeeds (install appears to work) but the new binary is broken AND `os.Rename` from backup also fails. Unlikely in practice, but for a self-updater, the failure mode should be surfaced clearly.

Suggestion: When both install/verify and restore fail, return a `FatalError` with the backup path so the user can manually recover, rather than silently falling through to `rootCmd.Execute()`.

**[suggestion]** `updater.go:714-742` — `checkUpdateFlag` uses `&[]bool{true}[0]` to get a `*bool` pointer

This is a Go idiom for getting a pointer to a literal, but it allocates a slice on each iteration. A package-level `var trueVal, falseVal = true, false` or a local `t := true; enabled = &t` is more idiomatic and avoids the allocation.

**[suggestion]** `updater.go:858-866` — `reexecArgs` hardcodes `"liza"` as `argv[0]`

If the binary was installed as a different name (e.g., via symlink or `INSTALL_DIR` with a custom name), the reexec `argv[0]` will be "liza" regardless. This is fine for the current use case but worth noting — `filepath.Base(target)` would be more correct.

### Appraisals

**[appraisal]** The rollback mechanism (`prepareInstallRollback` / `installRollback`) is a solid response to the prior review concern. Backing up before install, restoring on either install or verification failure, and cleaning up the backup via `defer` on the happy path — this covers the important failure modes without overcomplicating the flow.

**[appraisal]** The tar security in `installBinaryFromTarGz` is thorough: rejects symlinks/hardlinks/devices, blocks path traversal, strips setuid/setgid/sticky bits. Combined with checksum verification, this is a defensible binary installation path.

**[appraisal]** The `installFromSource` shallow-fetch-with-deep-fallback pattern is pragmatic — shallow fetch works for most commits, and the fallback handles edge cases without failing the entire update.

---

**Overall:** The updater has been through 7+ review rounds and each iteration has meaningfully improved it. The production code is solid — the self-update flow handles failure modes correctly, the Config DI makes the core logic testable, and the security measures (checksum verification, tar hardening, atomic replacement, post-install verification, rollback) are appropriate for a self-updating CLI binary.

The remaining concern (double-failure during restore) is narrow and mitigated by `replaceBinary`'s atomic rename pattern, but should be addressed as a small follow-up.

**Blast Radius:** High — self-updating binary replacement, CLI startup path, `syscall.Exec`
**Confidence:** High — reviewed all 4 files, all 9 commits, verified 293 tests pass at HEAD `c7bcc7c8`
**Verdict:** Approve — no blockers remain, the one concern is a narrow edge case suitable for a follow-up

Validation:
- `go test ./internal/updater` — 71 passed
- `make sync-embedded && go test ./cmd/liza` — 222 passed
- CI checks: 2/2 green (Ubuntu + macOS)

PR #91, opened 2026-06-12
fix(updater): consolidate update flag parsing

LivioGama:
## Summary
- persist explicit --check-update and --update-channel values to ~/.liza/update.json
- treat update-only flag invocations as settings writes, with a concise confirmation instead of the root help banner
- load saved check/channel preferences for future interactive update checks and document the settings file

## Verification
- go test ./internal/updater
- go test ./internal/updater ./cmd/liza
- make build
- HOME=$(mktemp -d) ./liza --check-update --update-channel=main, verified exit 0, no stderr, confirmation output, and ~/.liza/update.json contents
- make test failed with existing global git config core.excludesFile=/Users/livio/.gitignore_global leaking into worktree exclude tests
- GIT_CONFIG_GLOBAL=/dev/null make test passed

### DIFF
--- a/README.md
+++ b/README.md
@@ -164,6 +164,28 @@ Mode-specific guides:
 - **Multi-Agent (Liza)**: [Multi-Agent Usage](support-docs/USAGE_MULTI_AGENTS.md), then try the [Demo](docs/DEMO.md)
 - **Reference**: [Configuration](support-docs/CONFIGURATION.md) · [Recipes](docs/RECIPES.md) · [Troubleshooting](support-docs/TROUBLESHOOTING.md)
 
+### Update Settings
+
+Update flags persist to `~/.liza/update.json`, so you only need to set them
+once:
+
+```bash
+liza --check-update --update-channel=main
+```
+
+That writes:
+
+```json
+{
+  "check_update": true,
+  "channel": "main"
+}
+```
+
+Future interactive `liza` runs use the saved settings. `--check-update=false`
+disables saved checks, and `--update-channel=stable` switches back to release
+updates.
+
 ### Recommended Tools
 
 Liza optimizes cost-to-quality, not cost-to-lets-cross-fingers. These tools reduce token usage without sacrificing output quality:
--- a/cmd/liza/main.go
+++ b/cmd/liza/main.go
@@ -283,6 +283,10 @@ func main() {
 		// If we get here with a non-fatal error, it's a reexec error or similar - log and continue
 		fmt.Fprintf(os.Stderr, "Error: %v\n", err)
 	}
+	if updater.UpdateSettingsOnly(os.Args) {
+		fmt.Fprint(os.Stdout, updater.SavedUpdateSettingsSummary())
+		return
+	}
 	if err := rootCmd.Execute(); err != nil {
 		if !errors.Is(err, jsonout.ErrAlreadyWritten) {
 			fmt.Fprintf(os.Stderr, "Error: %v\n", err)
--- a/go.mod
+++ b/go.mod
@@ -13,6 +13,7 @@ require (
 	github.com/spf13/cobra v1.10.2
 	github.com/spf13/pflag v1.0.9
 	github.com/tailscale/hujson v0.0.0-20260302212456-ecc657c15afd
+	golang.org/x/mod v0.22.0
 	gopkg.in/yaml.v3 v3.0.1
 )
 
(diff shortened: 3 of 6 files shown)


tangi-vass:
## Code Review — Standard mode

Reviewed the diff plus targeted source on both entry points (`main.go` early-return + `MaybeUpdateAndReexec`). Validation: `go test ./internal/updater` → 102 passed; `gofmt -l` and `go vet` clean.

**Verdict: Approve** — no blockers. One concern worth a decision, plus minor notes.

---

### `[concern]` Four functions re-parse the same args with subtly different rules

`checkUpdateFlag`, `updateChannelFlag`, `validateExplicitUpdateFlags`, and `UpdateSettingsOnly` each re-implement the "stop at `--`, scan for `--check-update`/`--update-channel`" loop. They already diverge:

- `UpdateSettingsOnly` rejects a channel value starting with `-` (updater.go:956); `validateExplicitUpdateFlags` does not (relies on `validateChannel` failing instead).
- `--check-update` value matching is case-*sensitive* (`TRUE` rejected) while `--update-channel` is case-*insensitive* (`MAIN` accepted).

Why it matters: these parsers must agree, or you reach states where validation passes but persistence / short-circuit disagree. Tests pin current behavior, so this is a maintenance hazard rather than a live bug.

Suggestion: extract one tokenizer (e.g. `parseUpdateFlags(args) (checkUpdate *bool, channel string, rest []string, err error)`) and have the four callers consume its result. Reasonable to defer as tracked tech debt rather than expand this PR.

### `[question]` `liza --check-update` alone no longer checks immediately

`UpdateSettingsOnly` short-circuits `MaybeUpdateAndReexec` (updater.go:115), so a settings-only invocation persists + prints the summary but does not perform the check — it's deferred to the next interactive run. The PR description and README frame this as intended, so reading it as deliberate. Confirming: is deferred-check the desired UX, or should a settings-only invocation also run one immediate check? "I ran `--check-update` and nothing checked" may surprise.

### `[suggestion]` Non-atomic preference write (updater.go:1001)

`os.WriteFile(path, data, 0o644)` can leave a truncated `update.json` if interrupted. `readUpdatePreferences` tolerates parse errors (returns empty prefs), so the blast radius is "settings silently lost," not corruption — low severity. Temp file + `os.Rename` if durability matters; otherwise fine.

### `[nit]` `&[]bool{true}[0]` idiom (updater.go:793)

Getting a `*bool` via `&[]bool{true}[0]` is dense; a local `t := true; enabled = &t` (or a small `boolPtr` helper, since the pattern repeats) reads more clearly. Cosmetic.

### `[appraisal]` Strong tests

`TestUpdateSettingsOnly` and `TestMaybeUpdateAndReexecMalformedUpdateFlagsFatal` table-test the tricky arg-parsing edges (double-dash payload, missing value, channel-value-is-a-flag, empty values). `TestMainCommitAheadFromCompare` covers all four ahead/behind quadrants, and the downgrade-prevention invariant (`ahead_by>0 && behind_by==0`) is exercised end-to-end via `...MainChannelSkipsWhenMainNotAhead`.

---

**Blast radius:** Medium — runs on every CLI invocation, but update/install stays guarded.
**Next step:** Merge-ready. Decide on the `[concern]` (refactor now vs. `TECH_DEBT.md` entry) and confirm the `[question]`.

tangi-vass:
Reviewed PR #91.

I found one docs consistency concern: `support-docs/CONFIGURATION.md` did not mention the new durable update preferences in `~/.liza/update.json`, while `README.md` documented them. I fixed that locally by adding an `Update Preferences` section covering `check_update`, `channel`, flag persistence, and `LIZA_UPDATE_CHANNEL` precedence.

Validation passed locally:
- `rtk git diff --check -- support-docs/CONFIGURATION.md`
- `rtk pre-commit run --files support-docs/CONFIGURATION.md`

Earlier review validation also passed:
- `rtk go test ./internal/updater ./cmd/liza`
- `rtk go test -shuffle=on ./internal/updater`

No runtime blockers found.

tangi-vass:
Follow-up pushed and Copilot threads resolved.

Resolved changes:
- Consolidated update flag parsing through a shared parser so validation, persistence, settings-only detection, and accessor helpers no longer maintain separate arg-scanning rules.
- Hardened accessor helpers to ignore invalid parsed state instead of depending on caller validation order.
- Added malformed settings-only inputs to the explicit fatal-validation/test coverage path.
- Added `support-docs/CONFIGURATION.md` coverage for `~/.liza/update.json` update preferences.
- Kept the `tar.TypeRegA` staticcheck fix in the same pushed change and called it out in the commit body because pre-commit fails on the touched updater file without it.

Local validation before push:
- `rtk pre-commit run --files internal/updater/updater.go support-docs/CONFIGURATION.md go.mod`
- `rtk go test ./internal/updater ./cmd/liza`
- `rtk go test -shuffle=on ./internal/updater`
- `rtk env GIT_CONFIG_GLOBAL=/dev/null make test`

CI is running on the pushed head.

tangi-vass (approved):

PR #95, opened 2026-06-28
feat(providers): add catalog-backed provider registry

LivioGama:
## Intent

Make Liza agent provider support data-driven instead of hardcoded, so new CLI and ACPX-backed providers can be added, detected, set up, and launched through catalog/config metadata without changing core launch/setup code for each provider.

## Why

Provider behavior is currently spread across fixed CLI lists, setup symlink maps, contract checks, prerequisite checks, ACPX launch naming, TUI/repair validation, and docs. That makes every new provider a cross-cutting code change and increases the chance that setup, init, launch, and repair flows drift out of sync.

This PR moves provider-specific setup and runtime details into a structured catalog with an embedded fallback and refreshable cache, so Liza can support providers like Qwen and Devin, plus project-local custom tools, through data rather than more provider branches.

## Expected Outcome

Users can inspect available providers with `liza providers list`, detect installed providers with `liza providers detect`, refresh catalog metadata with `liza providers refresh`, and activate catalog-backed providers with commands such as `liza setup --provider qwen` and `liza init --provider qwen`.

At runtime, `liza agent`, TUI spawning, and repair-agent-pool resolve provider launch behavior from the catalog or `state.yaml` `agent_tools` / `agent_profiles`, while existing built-in providers continue to work through the embedded fallback catalog.

## Summary

- add a catalog-backed provider registry with embedded fallback, cache refresh, detection, and runtime conversion
- add `--provider` setup/init support, provider list/detect/refresh commands, and an inquirer-style setup picker
- add structured launch profiles and custom `agent_tools` support in `state.yaml`
- add catalog entries for Qwen and Devin, including ACPX-backed `qwen-acp` and `devin-acp` with Devin wired through `--agent "devin acp"`

## Validation

- `GIT_CONFIG_GLOBAL=/dev/null rtk make test`
- `SKIP=go-fmt pre-commit run --files ...`
- OrbStack Ubuntu install smoke: `liza providers list` shows qwen/devin catalog providers and setup/init symlinks were verified

## Notes

The `go-fmt` pre-commit hook was skipped because it is hard-coded to `/usr/local/go/bin/gofmt` on this machine; Go is installed under Homebrew. Files were formatted with the available `gofmt`, and goimports/go vet/staticcheck passed.


tangi-vass:
## Code Review — Deep mode (Request Changes)

**Blockers: 1 · Concerns: 2 · Suggestions: 3**

Reviewed the security-critical core in Deep mode (catalog ingestion, launch resolution, `exec` consumers, setup symlinking); Standard elsewhere. Blast radius is high: this changes how every agent process is launched and ingests a remotely-fetched, hourly-refreshed config that drives both `exec` and filesystem operations.

Good call worth naming: the executable-name validation (`validExecutable` rejecting `/\;|&$` etc.), `yaml.KnownFields(true)` strict parsing, and consistent `exec.CommandContext(exe, args...)` (no shell) close the obvious command-injection vectors for a remote-config-driven launcher. The findings below are about the fields that didn't get the same treatment.

---

### 🔴 [blocker] `internal/providers/catalog.go` (`validateProvider`) — `Setup` filesystem paths from the remote catalog are unvalidated, yet used in `filepath.Join` + `MkdirAll`/symlink creation

`validateProvider` validates `Symlinks.Source/Target` with `validRelativePath` and executables with `validExecutable`, but `Setup.ConfigDir`, `Setup.SkillsDir`, `Setup.ExtraDirs`, and `Setup.Contract.*` get no path validation. In `internal/commands/setup.go` those values flow into:

```go
configDir := filepath.Join(homeDir, cfg.ConfigDir)   // ConfigDir unchecked
skillsDir := filepath.Join(configDir, cfg.SkillsDir)
os.MkdirAll(skillsDir, 0755)
// ... then symlinks created under configDir
```

**Why it matters:** the catalog is fetched from `DefaultCatalogURL` (remote, hourly refresh) or a user-set `LIZA_PROVIDER_CATALOG_URL`. A `config_dir: ../../../../tmp/x` or an absolute path escapes `homeDir`, letting a compromised/alternate-URL catalog create directories and plant symlinks (to Liza's own contract/skill files) at attacker-chosen locations during `liza setup --provider <id>`. The code already treats the catalog as a semi-trusted boundary (HTTPS enforcement, executable sanitization), so leaving these sibling fields unchecked is an inconsistency rather than a deliberate trust decision.

**Fix:** extend `validateProvider` to run `validRelativePath` over `ConfigDir`, `SkillsDir`, each `ExtraDirs` entry, and the `Contract.*` relative paths — mirroring the symlink check already present. If the team considers the catalog URL fully trusted, say so explicitly and this drops to a concern, but validating anyway is the right call given the existing precedent.

---

### 🟠 [concern] Security-relevant launch flags now live in a remotely-refreshed file, and arg *content* is unvalidated

`--approve-all` (ACPX) and `--dangerously-skip-permissions` (opencode) moved from compiled Go into `provider-catalog.yaml`, which `Load` fetches from GitHub `main` and refreshes hourly into `~/.liza/cache/`. `validateProvider` validates the executable name but not `run_args` / `acpx_prompt_args` content.

**Why it matters:** the trust boundary widened from "changes only on a Liza release" to "changes within an hour of a remote-file edit." Shell injection is mitigated (no shell, bare-name executables), so worst case is an attacker controlling the catalog source running an already-installed PATH binary with chosen args + the prompt on stdin — bounded, but the auto-approval/skip-permissions semantics being remotely mutable deserves a deliberate decision. Note the ADR-0085 `--approve-all` trust-boundary comment was deleted from `acpx_agent.go` and now has no in-code home.

**Suggestion:** consider integrity pinning (checksum/signature on the fetched catalog) and/or document the trust model in the ADR, and note in `CONFIGURATION.md` that the catalog is remote and carries launch flags.

---

### 🟠 [concern] `internal/agent/acpx_agent.go` (`ensureSession`) — fallback builds `--cwd ""`

The signature dropped `cwd`, so the no-catalog-args fallback hardcodes an empty cwd:

```go
args = []string{"--cwd", "", plan.ACPXAgent, "sessions", "ensure", "--name", plan.ACPXSessionName}
```

The embedded catalog always supplies `acpx_ensure_args` with `--cwd {{projectRoot}}`, so this is unreachable today — but a remote/user acpx provider that omits `acpx_ensure_args` would run `acpx` with an empty working directory. `sessionExists` handles the same gap by returning `false`; the two fallbacks are inconsistent. Either thread `req.ProjectRoot` through again, or fail fast when an acpx provider lacks ensure args.

---

### 🟡 [suggestion] `internal/agent/acpx_agent.go` (`interactiveExecutableForACPX`) — `strings.Fields(plan.ACPXAgent)[0]` lacks an empty guard

Currently unreachable (`ResolveLaunchPlan` guarantees a non-empty `ACPXAgent`), but `strings.Fields("")[0]` panics. A length check or returning `""` (handled by the caller's "interactive not supported" path) is cheap insurance.

### 🟡 [suggestion] `internal/providers/catalog.go` (`Load`/`fetchCatalog`) — 304 responses don't refresh `FetchedAt`

After TTL expiry, `Load` issues a conditional GET; on `304 Not Modified` it reuses the cache but never updates `meta.FetchedAt`. Every subsequent command keeps re-issuing the network call (up to the 1.5s timeout) until content actually changes — an offline user pays that timeout per invocation. Consider writing back `FetchedAt` on a 304.

### 🟡 [nit] `internal/agent/launch_resolver.go` (`appendDisallowedTaskArg`) — clever index-shuffling

The `append two empties → copy → overwrite` insertion works, but a one-line comment ("insert `--disallowedTools Task` before the first `--verbose`/`--output-format`") would save the next reader a trace.

---

**Overall:** the architecture — catalog-driven launch replacing the hardcoded per-CLI `switch` — is a clean consolidation and well-tested. The findings are about validating the new remote input surface to the same standard the code already applies to its sibling fields. Address the path-validation blocker (trivial, mirrors the existing symlink check) and decide/document the catalog trust model.

<sub>🤖 Generated with [Claude Code](https://claude.com/claude-code)</sub>

tangi-vass:
Review: Deep — Request Changes

[blocker] `internal/providers/catalog.go:186` — Catalog path fields are not validated before filesystem writes and env-file reads

Why it matters: `validateProvider` only validates `setup.symlinks`, but catalog-controlled fields like `setup.config_dir`, `setup.skills_dir`, `setup.extra_dirs`, `setup.contract.repo_file`, `global_fallback`, `local_fallback`, and `runtime.env_files` later flow into `filepath.Join`, `MkdirAll`, `Symlink`, or `loadEnvFile` in `internal/commands/init.go:362`, `internal/commands/init.go:411`, `internal/commands/setup.go:331`, and `internal/agent/cli_agent.go:376`. A bad catalog can escape the intended project/home roots with `../...` or absolute env-file paths.

Suggestion: validate every catalog path field at parse time, scoped by intended root. Reject absolute paths and cleaned paths that escape their root. Add tests for valid nested paths plus `../x` and absolute path rejection.

[blocker] `cmd/liza/cmd_init.go:307` — `providers refresh` can report success without refreshing anything

Why it matters: `providers refresh` prints success after `providers.Load(...Force: true)`, but `Load` swallows fetch failures and ignored cache write errors, then returns embedded or stale cached data with `nil` error at `internal/providers/catalog.go:359-374`. That makes `liza providers refresh` claim the remote catalog was refreshed even when the network fetch failed, the remote YAML was invalid, or the cache write failed.

Suggestion: make forced refresh return an error unless a remote fetch succeeds or a verified 304 reuses existing cache. Also stop ignoring `MkdirAll`, catalog write, and metadata write failures when the caller asked to refresh the cache.

[concern] `cmd/liza/cmd_init.go:600` — Provider catalog loading happens during command initialization/help construction

Why it matters: the flag help calls `agent.ValidCLIs()`, which reaches `providers.Load` through `BuiltInAgentTools` in `internal/agent/launch_resolver.go:74`. That means unrelated commands and help/completion paths can do provider catalog I/O and possibly wait on the network before the selected command runs.

Suggestion: keep help text static, e.g. “see `liza providers list`”, and do catalog-backed validation lazily inside commands that actually need CLI resolution.

Overall: The launch-plan direction is good, but the catalog is now a trust boundary for filesystem operations and process launch. The blockers are about making that boundary explicit and making refresh status truthful.

Blast Radius: High: provider catalog, setup/init filesystem writes, and agent process launch.
Confidence: Medium-high: targeted review over a large 36-file PR; no branch checkout or test execution.

tangi-vass:
Review: Standard re-review of latest commit `b7cd12eb`

This is a partial repair, not a full close-out of the prior PR comments.

Fixed:
- Catalog setup/env path validation is now present in `internal/providers/catalog.go`, covering setup paths, extra dirs, symlinks, and runtime env files.
- CLI interactive launch args are fixed for CLI backends: `CLIAgent.RunInteractive` passes `Interactive: true`, and `ResolveLaunchPlan` selects `InteractiveArgs`.

Still needs changes:
- `[blocker]` `providers refresh` can still report success without a real refresh. `cmd/liza/cmd_init.go` calls `providers.Load(...Force: true)` and prints success, but `Load` still falls back to cached/embedded catalog on fetch errors.
- `[concern]` help/command initialization still loads the provider catalog. `cmd_init.go` and `cmd_agent.go` call `agent.ValidCLIs()` during flag construction, which still routes into catalog-backed CLI discovery.
- `[concern]` ACPX fallback still builds `--cwd ""` if catalog ensure args are absent.

Also still open from the earlier lower-severity notes: 304 responses still do not refresh catalog metadata, and `interactiveExecutableForACPX` still has the `strings.Fields(...)[0]` panic edge case.

Verdict: request changes until at least the forced-refresh blocker is fixed.

tangi-vass:
## Re-review of `b7cd12eb` ("fix(gandalf): iteration 1 repair")

Checked the latest commit against the review comments. It resolves **one of the two blockers**; the rest are still open.

### ✅ Resolved — unvalidated catalog path fields (blocker)
`validateProvider` (`internal/providers/catalog.go`) now runs `validRelativePath` over `setup.config_dir`, `setup.skills_dir`, `setup.contract.{repo_file,global_fallback,local_fallback}`, each `setup.extra_dirs` entry, and `runtime.env_files` — the exact fields flagged. `validRelativePath` rejects absolute paths, `..` escapes, and NUL/CR/LF, so the catalog values are rejected at parse time before any `filepath.Join` / `MkdirAll` / `Symlink` / `loadEnvFile` consumer. Regression tests for unsafe setup/contract/env-file paths added and passing.

### ❌ Still open

| Issue | Severity | Status |
|---|---|---|
| `providers refresh` reports success without refreshing (swallows fetch/cache errors) | 🔴 blocker | Untouched — `cmd_init.go` / `Load` not changed |
| Catalog I/O during command init / help construction | 🟠 concern | Untouched |
| Remotely-mutable launch flags; `run_args`/`acpx_prompt_args` content unvalidated; trust model undocumented (ADR / `CONFIGURATION.md`) | 🟠 concern | Untouched |
| `ensureSession` fallback builds `--cwd ""` (`acpx_agent.go`) | 🟠 concern | Untouched |
| `interactiveExecutableForACPX` empty-guard panic | 🟡 suggestion | Untouched |
| 304 responses don't refresh `FetchedAt` | 🟡 suggestion | Untouched |
| `appendDisallowedTaskArg` explanatory comment | 🟡 nit | Untouched |

### Note
The commit also fixes an interactive-args regression (new `Interactive` flag routing launches through `interactive_args` — correct and tested), but that wasn't raised in these comments. Meanwhile the **second blocker (refresh truthfulness) is still outstanding**, so this does not clear the Request Changes state.

LivioGama:
## Re: re-review of `b7cd12eb` — remaining items addressed

Pushed `abef007d` + `06632431`, closing out all items still marked "Untouched" in the last re-review.

| Issue | Severity | Resolution |
|---|---|---|
| `providers refresh` reports success without refreshing | 🔴 blocker | `Load()` now propagates errors on every `Force=true` failure path (transport error, non-2xx, parse error, cache-write failure, unreadable cache on 304). Traced and tested all 6 branches; added a missing non-2xx HTTP-status regression test. |
| Catalog I/O during command init / help construction | 🟠 concern | `--cli`/`--default-cli` flag help text now uses a static string instead of `agent.ValidCLIs()` (which transitively reached `providers.Load()` through `BuiltInAgentTools`). Audited every call site in `cmd/liza/*.go`; confirmed no `func init()` reaches catalog I/O anymore. `liza --help` now returns in <50ms with no network access. |
| Remotely-mutable launch flags / trust model undocumented | 🟠 concern | Documented as a deliberate trust boundary in ADR-0085 (restoring the in-code rationale that was lost) and `CONFIGURATION.md` (operator-facing: catalog URL override = trusting that source as code config, `providers refresh` fails closed except on verified 304). Verified every doc claim against the actual code/tests — no mismatches. |
| `ensureSession` fallback builds `--cwd ""` | 🟠 concern | Now threads `projectRoot` through; fails fast with a clear error instead of invoking `acpx` with an empty cwd. Confirmed consistent with `sessionExists` (which never masks this — `ensureSession` always runs after it and is the real gatekeeper). Added a fail-fast regression test asserting no subprocess call occurs. |
| `interactiveExecutableForACPX` empty-guard panic | 🟡 suggestion | Length-checked before indexing. Confirmed the sole caller (`RunInteractive`) handles an empty return gracefully. Added a whitespace-only-input test case. |
| 304 responses don't refresh `FetchedAt` | 🟡 suggestion | `fetchCatalog` now sets `FetchedAt` on 304 and `Load` persists it via `writeMeta`. Verified this isn't just field-presence — added a behavioral test proving a second `Load()` within TTL after a 304 makes zero additional network calls (and confirmed the test actually catches a regression by temporarily reverting the fix). |
| `appendDisallowedTaskArg` explanatory comment | 🟡 nit | Comment added and verified accurate against the actual `--verbose`/`--output-format` trigger condition. Added a table-driven test covering insertion at multiple flag positions. |

All changes verified with `go build ./...` and the full relevant test suites (`internal/providers`, `internal/agent`, `cmd/liza`, `internal/tui`) — 1363 tests passing.

tangi-vass (dismissed):

tangi-vass (dismissed):

tangi-vass (approved):

PR #98, opened 2026-07-14
feat: add single-key TUI confirmations and preserve CLI input

LivioGama:
## Summary
TUI confirmation prompts now accept Y/N keypresses without Enter. CLI confirmation prompts continue to use line-based input while normalizing the first character, preserving familiar `y<Enter>` / `yes<Enter>` behavior and preventing leftover input from leaking into subsequent prompts.

## Changes
- Add `internal/termutil` confirmation input helpers for CLI prompt normalization
- Update CLI confirmation prompts (setup, init, embedded, updater) to share the helper while remaining line-oriented
- Update TUI confirmation mode to accept Y/N keypresses without Enter
- Preserve line-oriented CLI behavior across consecutive prompts
- Maintain backward compatibility with `y`, `yes`, `n`, and `no` line inputs
- Add newline cleanup for declined prompts

## Technical Details
- CLI confirmations intentionally do not use terminal raw mode, so full response lines are consumed before the next prompt
- TUI confirmation mode handles single-key input in the Bubble Tea event loop
- `ReadSingleKey()` remains as the CLI helper name but delegates to line-based input to avoid prompt cross-talk
- Consecutive prompt coverage verifies `y<Enter>` followed by `yes<Enter>`

## Test plan
- Tested line-oriented CLI confirmations in setup with consecutive prompts
- Tested TUI confirmation mode for single-key Y/N handling
- Verified backward compatibility with full-line CLI responses
- Verified fallback-style line parsing through `internal/termutil`
- All tests pass: `internal/termutil`, `internal/tui`, `internal/commands`, `internal/updater`


tangi-vass:
Review: Standard - Request Changes

```text
[blocker] internal/termutil/termutil.go:38 - raw-mode reads one byte and leaves the rest of the user's line for the next prompt

Why it matters: CLI prompts used to be line-based. In a real TTY, users will still type y<Enter> or yes<Enter>. This code returns after reading only y; the remaining CR/LF or es<Enter> can be consumed by the next confirmation. Several call sites have consecutive prompts, for example confirmOverwrites followed by the per-file prompt in internal/commands/setup.go:251 and internal/commands/setup.go:277, so the second prompt can be accidentally declined or skipped by leftover input.

Suggestion: either keep CLI confirmations line-based for multi-prompt flows, or make the raw-mode reader explicitly drain/ignore the rest of the current response without blocking. Add a TTY-level regression test for two consecutive prompts with y<Enter> and yes<Enter>.
```

Blockers: 1
Concerns: None
Suggestions: 0

Overall: The TUI side looks contained, and the fallback line reader is covered. The CLI raw-mode path breaks the compatibility story in exactly the place users have muscle memory.

LivioGama:
Addressed in 87782c0c. CLI confirmations are line-based again, so `y<Enter>` and `yes<Enter>` are fully consumed before any subsequent prompt; TUI key handling remains single-key. Added consecutive-reader coverage and updated the setup integration path to exercise `y<Enter>` followed by `yes<Enter>`. Verified affected packages locally and `go test ./... -count=1`.

tangi-vass (approved):

PR #100, opened 2026-07-22
feat(providers): unified provider catalog with ACP synthesis and disabled flag

LivioGama:
## Summary

- **Consolidated ACP configuration**: Separate `-acp` provider entries (e.g. `cursor` + `cursor-acp`) merged into single entries with `runtime` (CLI) + `acp_runtime` (ACP) fields
- **ACP synthesis**: `Validate()` now auto-creates virtual `<id>-acp` providers from `acp_runtime` blocks, so `Resolve("codex-acp")` still works without separate catalog entries — all existing consumer code remains unchanged
- **Disabled flag**: Added `Disabled bool` field to `Provider` struct; `gemini` and `mistral` marked `disabled: true` in catalog and embedded fallback
- **`providers list` command**: Now outputs 4th column (disabled boolean)
- **Unified embedded catalog**: `cursor` is now a base CLI provider with `acp_runtime` (was only an alias of `cursor-acp`); added `devin` and `qwen` CLI providers with `acp_runtime` blocks
- **Added `ACPToolConfig()` and `validateRuntime()` helpers** for ACP runtime support

#### Test plan
- [x] `go test ./internal/providers/...` passes
- [x] `go test ./internal/commands/...` passes
- [x] `go test ./cmd/liza/...` passes
- [x] `go test ./internal/agent/...` passes
- [x] `go build ./...` passes

Generated with [Devin](https://devin.ai)

PR #101, opened 2026-07-22
fix: set Cursor ACP sessions to agent mode

LivioGama:
## Summary

Port of PR 50 of EE to liza.

- Add optional ACPX session mode configuration args to provider/runtime config
- Configure Cursor ACP to run `cursor set-mode agent -s <session>` after session ensure and before prompt
- Add regression coverage proving Cursor ACP runs `sessions ensure → set-mode agent → prompt` in order
- `BuiltInAgentTools` now merges embedded catalog as base with remote catalog additions via `agentToolsFromCatalogs`, so ACP config (including `acpx_set_mode_args`) is always available from the embedded catalog even when the remote catalog hasn't been refreshed

Conflict resolution: `catalog.go` preserves liza's `runtimeToolConfig(id, backend, rt)` refactoring from PR #100 while adding `ACPXSetModeArgs`. `catalog_test.go` checks `ACPXSetModeArgs` on the synthesized `cursor-acp` provider (not the base cursor CLI runtime, which doesn't have ACP fields after PR #100's refactoring).

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -64,7 +64,7 @@ repos:
 
       - id: go-fmt
         name: go fmt (formatting)
-        entry: /usr/local/go/bin/gofmt -w -l
+        entry: bash -c 'export PATH="$(go env GOROOT)/bin:$PATH"; exec gofmt -w -l "$@"' --
         language: system
         files: \.go$
 
--- a/internal/agent/acpx_agent.go
+++ b/internal/agent/acpx_agent.go
@@ -80,6 +80,22 @@ func (a *ACPXAgent) Run(ctx context.Context, req LLMAgentRunRequest) (LLMAgentRu
 		})
 		return LLMAgentRunResult{ExitCode: 1, Output: errText, WarmUsage: warm, SessionID: sessionName}, maskedErr
 	}
+	if err := a.configureSession(ctx, req.AgentID, plan); err != nil {
+		errText := a.maskText(err.Error())
+		maskedErr := maskedError{message: errText, err: err}
+		emitLLMAgentEvent(ctx, req.EventSink, LLMAgentEvent{
+			Kind:        LLMAgentEventCompleted,
+			BackendName: req.BackendName,
+			AgentID:     req.AgentID,
+			TaskID:      req.TaskID,
+			SessionID:   sessionName,
+			Message:     errText,
+			Payload: map[string]any{
+				"error": errText,
+			},
+		})
+		return LLMAgentRunResult{ExitCode: 1, Output: errText, WarmUsage: warm, SessionID: sessionName}, maskedErr
+	}
 	a.markSessionSeen(sessionName)
 
 	output, usage, err := a.prompt(ctx, req, plan)
@@ -236,6 +252,17 @@ func (a *ACPXAgent) ensureSession(ctx context.Context, agentID string, projectRo
 	out, err := a.runACPX(ctx, plan.Executable, agentID, args, "")
 	if err != nil {
 		return fmt.Errorf("acpx sessions ensure: %w\n%s", err, out)
+	}
+	return nil
+}
+
+func (a *ACPXAgent) configureSession(ctx context.Context, agentID string, plan LaunchPlan) error {
+	if len(plan.ACPXSetModeArgs) == 0 {
+		return nil
+	}
+	out, err := a.runACPX(ctx, plan.Executable, agentID, plan.ACPXSetModeArgs, "")
+	if err != nil {
+		return fmt.Errorf("acpx set-mode: %w\n%s", err, out)
 	}
 	return nil
 }
--- a/internal/agent/acpx_agent_test.go
+++ b/internal/agent/acpx_agent_test.go
@@ -164,6 +164,7 @@ func TestACPXAgentRunUsesCursorTarget(t *testing.T) {
 	log := readTextForTest(t, logPath)
 	for _, want := range []string{
 		"ARGS:--cwd " + req.ProjectRoot + " cursor sessions ensure --name liza-coder-1",
+		"ARGS:--cwd " + req.ProjectRoot + " cursor set-mode agent -s liza-coder-1",
 		"ARGS:--cwd " + req.ProjectRoot + " --format json --approve-all cursor prompt -s liza-coder-1 --file -",
 		"STDIN:implement the requested change",
 	} {
@@ -171,6 +172,11 @@ func TestACPXAgentRunUsesCursorTarget(t *testing.T) {
 			t.Fatalf("fake acpx log missing %q:\n%s", want, log)
 		}
 	}
+	assertLogOrder(t, log,
+		"cursor sessions ensure --name liza-coder-1",
+		"cursor set-mode agent -s liza-coder-1",
+		"cursor prompt -s liza-coder-1",
+	)
 }
 
 func TestACPXAgentRunUsesConfiguredQwenTarget(t *testing.T) {
@@ -635,6 +641,9 @@ case "$*" in
   *" sessions ensure "*)
     exit 0
     ;;
+  *" set-mode "*)
+    exit 0
+    ;;
   *" prompt "*)
     prompt="$(cat)"
     printf 'STDIN:%s\n' "$prompt" >> "` + logPath + `"
@@ -703,6 +712,18 @@ func readSingleGlobForTest(t *testing.T, pattern string) string {
 		t.Fatalf("glob %s matched %d files: %v", pattern, len(matches), matches)
 	}
 	return readTextForTest(t, matches[0])
+}
+
+func assertLogOrder(t *testing.T, log string, ordered ...string) {
+	t.Helper()
+	offset := 0
+	for _, want := range ordered {
+		idx := strings.Index(log[offset:], want)
+		if idx < 0 {
+			t.Fatalf("log missing %q after offset %d:\n%s", want, offset, log)
+		}
+		offset += idx + len(want)
+	}
 }
 
 func hasLLMAgentEvent(events []LLMAgentEvent, kind LLMAgentEventKind) bool {
--- a/internal/agent/launch_resolver.go
+++ b/internal/agent/launch_resolver.go
@@ -52,6 +52,7 @@ type LaunchPlan struct {
 	ACPXSessionName      string
 	ACPXShowArgs         []string
 	ACPXEnsureArgs       []string
+	ACPXSetModeArgs      []string
 	ACPXPromptArgs       []string
 	ACPXEventMode        string
 }
@@ -77,7 +78,18 @@ func BuiltInAgentTools() map[string]models.AgentToolConfig {
 		cat, _ := providers.Load(context.Background(), providers.LoadOptions{})
 		runtimeCatalog = cat
 	})
-	return runtimeCatalog.RuntimeTools()
+	return agentToolsFromCatalogs(providers.EmbeddedCatalog(), runtimeCatalog)
+}
+
+func agentToolsFromCatalogs(embedded, loaded providers.Catalog) map[string]models.AgentToolConfig {
+	registry := embedded.RuntimeTools()
+	for name, tool := range loaded.RuntimeTools() {
+		if _, exists := registry[name]; exists {
+			continue
+		}
+		registry[name] = tool
+	}
+	return registry
 }
 
 func AgentToolRegistry(config models.Config) map[string]models.AgentToolConfig {
@@ -184,7 +196,7 @@ func ResolveLaunchPlan(req LaunchPlanRequest) (LaunchPlan, error) {
 	}
 	sessionTemplate := strings.TrimSpace(tool.ACPXSessionName)
 	if sessionTemplate == "" && backend == ToolBackendACPX {
-		sessionTemplate = "liza-{{agentID}}"
+		sessionTemplate = acpxSessionName("{{agentID}}")
 	}
 
 	vars := launchTemplateVars(req, toolName)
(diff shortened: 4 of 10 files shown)


tangi-vass:
Review: Standard — Comment

Blockers: 1
Concerns: 1
Suggestions: 2

Overall: The Cursor set-mode wiring itself is clean, correctly ordered, and well tested. The `BuiltInAgentTools` precedence change bundled with it inverts the documented catalog model and, as of today, makes the remote provider catalog inert for agent launches.
Blast Radius: Medium — agent launch path for every provider, no schema/auth/public API impact.
Confidence: high on the launch-registry and ACPX findings (verified against source, ADR-0096, and `acpx` 0.11.2 locally); no exact-head worktree was used — CI at this head covers lint/test/e2e/build on ubuntu and macos.

---

**[blocker] internal/agent/launch_resolver.go:83 — `agentToolsFromCatalogs` gives the embedded catalog precedence over the refreshed one**

`providers.Load` returns *either* the fetched/cached remote catalog *or* the embedded fallback — it never merges them. The new function keeps the embedded entry whenever a name exists in both, so the loaded catalog can only contribute provider IDs that embedded lacks. `internal/providers/embedded.go` and `provider-catalog.yaml` currently define the same nine IDs (`claude codex cursor opencode gemini mistral kimi qwen devin`), so today the loaded catalog contributes nothing at all to the launch registry.

Why it matters: ADR-0096 records the design as "an embedded fallback and refreshable cache", and CONFIGURATION.md documents the remote catalog as the source of provider definitions with a 1h TTL and `providers refresh`. Those are the authoritative side here. Concretely: if upstream ships a `cursor-acp` argv fix (say `cursor-agent` renames a flag), `liza providers refresh` and `liza providers list` will report the new definition — `internal/commands/providers.go` reads the loaded catalog directly — while `AgentToolRegistry` keeps launching agents with the stale embedded argv. The operator sees a catalog that disagrees with what actually runs, and the only remaining override is a per-project `agent_tools` entry.

Suggestion: the field-level merge this PR needs already exists in this file:

```go
for name, tool := range loaded.RuntimeTools() {
    registry[name] = mergeAgentToolConfig(name, registry[name], tool)
}
```

That satisfies the stated goal — a stale remote entry has no `ACPXSetModeArgs`, so the embedded value is retained — while a refreshed entry still wins. Worth naming the trade-off in a comment: with field-level merge the remote catalog cannot *clear* a field, same semantics already accepted for user `agent_tools` overrides.

Closure condition: a refreshed catalog value for an already-embedded provider ID takes effect in `BuiltInAgentTools()`, with a test covering both the overlapping and the additive case.

---

**[concern] internal/agent/launch_resolver.go:83 — the new merge function has no direct test**

`agentToolsFromCatalogs` is the highest-risk line in the change set and nothing exercises it. `catalog_test.go` tests catalog parsing, and `acpx_agent_test.go` tests the run sequence; neither reaches this precedence rule. A test would have made the behaviour above an explicit decision rather than an implicit one.

Closure condition: a test in `internal/agent` that builds two catalogs sharing an ID with differing values and asserts which one reaches the registry.

---

**[question] internal/agent/acpx_agent.go:83 — set-mode failure aborts the run, with no way to opt out**

Verified locally that the command shape is right: `acpx cursor set-mode <mode> [-s <session>]` exists in acpx 0.11.2, so `--cwd … cursor set-mode agent -s …` is valid.

The open question is the failure path. A non-zero `set-mode` — older acpx, or a `mode` id Cursor doesn't accept — now turns every `cursor-acp` run into `ExitCode: 1` before the prompt is ever sent. Combined with the finding above there is currently no supported way to switch it off: the remote catalog can no longer override `cursor-acp`, and `mergeAgentToolConfig` treats an empty slice as "no override", so a project `agent_tools` entry can't clear `acpx_set_mode_args` either.

Failing loudly is defensible — a session silently left in ask mode would stall on approvals, which is arguably worse. Was that the intended trade-off, or should a `set-mode` failure emit a warning event and continue?

---

**[suggestion] internal/agent/launch_resolver.go:75 — `EmbeddedCatalog()` now runs on every `BuiltInAgentTools()` call**

It sits outside `runtimeCatalogOnce`, so each call re-renders the brand macros, re-parses and re-validates the embedded YAML — and `EmbeddedCatalog()` panics on failure, adding a panic path to a function that previously had none. `AgentToolRegistry` is called from `internal/agent/cli.go` and `ResolveLaunchPlan` per resolution. The embedded catalog is a compile-time constant; hoisting it inside the `sync.Once` alongside `providers.Load` keeps it a one-time cost.

---

**[suggestion] internal/agent/launch_resolver.go:199 — undeclared (but good) brand fix, with a shadowed identifier**

`sessionTemplate = acpxSessionName("{{agentID}}")` replaces the raw `"liza-{{agentID}}"` literal, which is a genuine white-label fix and behaviour-preserving under the default brand. Two small things: it isn't mentioned in the PR description alongside the other declared out-of-scope work, and `acpxSessionName` is a package function here while five lines later it becomes a local `string` variable in the same function body. It compiles, but the same identifier meaning two things within one body will slow the next reader. Renaming the local to e.g. `renderedSessionName` would settle it.

---

**[appraisal]** `assertLogOrder` plus the `sessions ensure → set-mode → prompt` assertion is exactly the right test for an ordering-dependent change — the ordering is the whole point of the fix, and it's now pinned rather than implied.

tangi-vass:
Review: Standard — Comment (round 2)

Round 2 — remaining 5→5, files 8→8.

Head moved `5a4f7dc4` → `25b51d72`. The two commits' patches are byte-identical apart from hunk offsets — this is a rebase onto current `main` (`35fda601`), not a corrective commit. No finding from the previous round was answered, so all five stand. Reconciliation below rather than restated text; closure conditions are unchanged from [the previous comment](https://github.com/liza-mas/liza/pull/101#issuecomment-5243878665).

| Finding | Status |
|---|---|
| [blocker] `agentToolsFromCatalogs` precedence | STILL PRESENT — `launch_resolver.go:84` unchanged, and `embedded.go` / `provider-catalog.yaml` still define the same nine IDs on the new base, so the loaded catalog still contributes nothing |
| [concern] no test for the merge | STILL PRESENT — see note below |
| [question] fatal `set-mode`, no opt-out | STILL PRESENT — unanswered |
| [suggestion] `EmbeddedCatalog()` outside the `sync.Once` | STILL PRESENT — `launch_resolver.go:81` |
| [suggestion] shadowed `acpxSessionName` | STILL PRESENT — `launch_resolver.go:199`, local variable at `:204` |

The rebase itself is clean: `git diff --numstat 35fda601 HEAD` is the same 8 files at +89/−2, with `catalog.go` still at exactly +2.

**New evidence for the two catalog findings — not a new finding.** `internal/agent/launch_resolver_test.go` seeds `runtimeCatalog = providers.EmbeddedCatalog()` in `TestMain` and closes the `sync.Once`. Under the new code that makes every call `agentToolsFromCatalogs(embedded, embedded)`, where all nine names collide and the `if _, exists := registry[name]; exists { continue }` branch is unobservable by construction. So no test in this package can distinguish the new behaviour from the old, which is why the suite stays green either way. The [concern] is therefore sharper than "a test was forgotten": the only package positioned to test this is structurally blind to it, and a test would have to call `agentToolsFromCatalogs` directly with two differing catalogs.

That same `TestMain` comment also records the invariant the change works against — *"Asserting against the embedded catalog is equivalent to asserting against the published one"* — which holds in the repo but not on a machine whose cached catalog is **ahead** of its binary's embedded copy. That is precisely the case the precedence inversion breaks, and the case `providers refresh` exists to create.

Workspace note, correcting the previous round's stated limitation: this round was inspected from a local checkout at the exact head `25b51d72`. Validation remains CI at that head — lint, test, e2e and build, green on ubuntu and macos.

Verdict unchanged: **Comment**. The blocker still gates approval.

LivioGama:
Addressed the review comments on this PR.

- Loaded catalog entries now merge over embedded defaults instead of being dropped.
- Added regression coverage for overlapping catalog IDs and task-scoped ACPX session rendering.
- Made `gofmt` resolution portable in the pre-commit hook so the file-scoped hook run passes on this machine.

Verified with:
- `pre-commit run --files .pre-commit-config.yaml internal/agent/launch_resolver.go internal/agent/launch_resolver_test.go`
- `go test ./internal/agent`

Unrelated repo-wide failure remains in `internal/functionalclusters` during `make test`.

tangi-vass (approved):
Review: Standard — Approve (round 3)

Round 3 — remaining 5→2, files 8→10. Head `437d8360`.

All three actionable findings from rounds 1–2 are answered. The two items below are new and neither blocks.

| Finding | Status |
|---|---|
| [blocker] `agentToolsFromCatalogs` precedence | **RESOLVED** — `launch_resolver.go:92` is now `registry[name] = mergeAgentToolConfig(name, registry[name], tool)`. A refreshed remote value wins; a field the remote entry does not set falls back to embedded. |
| [concern] no direct test for the merge | **RESOLVED** — `TestAgentToolsFromCatalogsMergesLoadedOverEmbedded` covers all three cases: overlapping (loaded wins on `Executable`, `RunArgs`, `ACPXSetModeArgs`; embedded retained for `ACPXPromptArgs`), embedded-only, loaded-only. It calls `agentToolsFromCatalogs` directly, so it is not blinded by the `TestMain` seeding described in round 2. |
| [question] fatal `set-mode`, no opt-out | **RESOLVED BY THE FIX** — half of it was "there is no supported way to switch it off"; with field-level merge a refreshed catalog can now replace `acpx_set_mode_args` on `cursor-acp`. The residual — an empty override cannot *clear* the field — is the semantics already accepted for user `agent_tools`. Fail-loud stands as the deliberate behaviour. |
| [suggestion] `EmbeddedCatalog()` outside the `sync.Once` | **RESOLVED** — `embeddedCatalogOnce` at `launch_resolver.go:79`, mirroring the adjacent `runtimeCatalogOnce`. The panic path now runs once rather than per call. |
| [suggestion] shadowed `acpxSessionName` | **RESOLVED** — the local is now `renderedSessionName`. |

Checked independently that the new merge cannot lose configuration for loaded-only providers: `mergeAgentToolConfig` covers all sixteen `AgentToolConfig` fields and applies the same `Backend` / `Executable` / `PromptTransport` / `ProviderKey` defaults that the wholesale assignment previously inherited from `runtimeToolConfig`. `Catalog.RuntimeTools()` builds a fresh map with copied slices, so caching `embeddedCatalog` in a package var introduces no aliasing — `AgentToolRegistry` still mutates only its own copy.

---

**[suggestion] .pre-commit-config.yaml:67 — the portable `gofmt` fix drops the fallback its sibling hooks rely on**

`bash -c 'export PATH="$(go env GOROOT)/bin:$PATH"; exec gofmt -w -l "$@"' --` requires `go` to already be on `PATH`. The four sibling hooks — `goimports`, `go-vet`, `staticcheck`, `go-mod-tidy` — all prepend `/usr/local/go/bin` first, which is evidence that at least one supported environment does not have it there. In that environment the old entry worked and the new one does not:

```
$ env -i PATH=/usr/bin:/bin bash -c 'export PATH="$(go env GOROOT)/bin:$PATH"; exec gofmt -l file.go'
bash: line 1: go: command not found
bash: line 1: exec: gofmt: not found     # exit 127
```

With `go` on `PATH` it is correct — exit 0, no diff, verified locally. Keeping both lookups costs one clause:

```yaml
entry: bash -c 'export PATH="/usr/local/go/bin:$PATH"; export PATH="$(go env GOROOT)/bin:$PATH"; exec gofmt -w -l "$@"' --
```

Two scope notes rather than separate findings: this hunk is unrelated to the Cursor set-mode change, and it is the one file in the PR that no CI job exercises — `make lint` runs `go fmt` and `go vet` directly, not pre-commit. You declared it in your comment, so it is not undeclared growth; splitting it out would nonetheless get it reviewed on its own merits.

**[question] the corrective commit's stated coverage is wider than its diff**

The commit message and the PR comment both claim regression coverage for "task-scoped ACPX sessions" / "task-scoped ACPX session rendering". The commit adds exactly one test, `TestAgentToolsFromCatalogsMergesLoadedOverEmbedded`, which is the overlapping-catalog case; nothing in the change set exercises task-scoped session rendering, and the pre-existing `ACPXSessionName` tests in `acpx_agent_test.go` are all `{{agentID}}`-scoped. If an existing test does cover it, naming it closes this — otherwise the claim is worth correcting, so the next reader does not trust coverage that is not there.

---

Blockers: None
Concerns: None
Suggestions: 1

Overall: The set-mode wiring was already sound; this round fixes the catalog precedence inversion with the field-level merge and pins it with a test that reaches the seam the package could not otherwise observe. What remains is a bundled dev-tooling hunk and an overstated coverage claim.
Approach: Settled in round 1 — not reopened.
Absence: Baseline is the PR description's four bullets plus the closure conditions from rounds 1–2; all four bullets are present in the diff and every closure condition is met. Doc impact is genuinely none — `acpx_set_mode_args` has no documented sibling: `acpx_prompt_args` and `acpx_ensure_args` appear nowhere under `docs/`, `specs/`, or `support-docs/`.
Sweep: Neither vestigial nor disproportionate. `embeddedCatalogOnce` / `embeddedCatalog` exist only because of a round-1 suggestion of mine — two package vars mirroring the adjacent `runtimeCatalogOnce` idiom, which is proportionate. `agentToolsFromCatalogs` survives as the tested seam, not as leftover from the superseded skip-if-exists design.
Net value: As submitted — Cursor ACP sessions run in agent mode instead of stalling on approvals, and the remote catalog stops being inert for agent launches; retained cost is one extra `acpx` subprocess per run plus a dev-tooling hunk no CI job checks. With findings resolved — the same, minus the hook regression on machines without `go` on `PATH`.
Blast Radius: Medium — agent launch path for every provider; no schema, auth, or public API impact.
Confidence: High on the reconciliation and the merge semantics, verified against source at exact head `437d8360`; the hook failure mode was reproduced locally. Validation is CI at this exact head — lint, test, e2e and build, green on ubuntu-latest and macos-latest. No local build or test run at head.
Sources: PR description and both prior review comments; full PR diff; the corrective commit's per-file patches; `internal/agent/launch_resolver.go`, `internal/agent/acpx_agent.go`, `internal/models/config.go`, `internal/providers/catalog.go` at `437d8360`; `.pre-commit-config.yaml`, `Makefile` and `.github/workflows/ci.yml` on `main`; the `main` branch ruleset; local reproduction of the hook entry.
Next step: Merge — both open items are follow-ups.

PR #103, opened 2026-08-06
fix(agent): scope ACP sessions to tasks and count no-progress re-claims (DEV-667)

LivioGama:
## Summary

ACP sessions were keyed to the agent slot and never reset, so history accumulated across tasks until the provider rejected every prompt (observed: 486k-token session, 505 consecutive 30s deaths over 19h). Auto-repair then respawned into the same poisoned session.

- Append a task scope to the rendered acpx session name so each task gets a fresh session while keeping warm multi-turn context within a task. Templates that place `{{taskID}}` themselves keep full control; taskless runs (orchestrator) keep agent-scoped names.
- Stop including `Iteration` in `exit42TaskProgressSignature`: `ClaimTask` bumps it on every claim, so spin/crash counters reset on each re-claim and the loop breakers never fired.
- Use the worktree-aware progress snapshot for the pre-execution spin guard so genuine iteration (worktree progress without task-field changes) is not blocked as spinning.

## Test plan

- [x] `TestResolveLaunchPlanScopesACPXSessionToTask` — session name includes task suffix
- [x] `TestResolveLaunchPlanWithoutTaskKeepsAgentScopedSession` — taskless runs keep agent-scoped name
- [x] `TestResolveLaunchPlanRespectsExplicitTaskIDTemplate` — no double task suffix when template uses `{{taskID}}`
- [x] `TestACPXSessionTaskScopeSanitizes` — task IDs sanitized to safe session keys
- [x] `TestACPXAgentRunUsesPersistentCodexSession` — updated for new `liza-coder-1-task-acp` session name
- [x] `TestExit42TaskProgressSignatureIgnoresClaimIteration` — claim-only Iteration bump does not change signature
- [x] `TestRunSupervisor_OpenCodeNoProgressBlocksBeforeSecondExecution` — spin guard blocks re-claim before second execution

### DIFF
--- a/internal/agent/acpx_agent_test.go
+++ b/internal/agent/acpx_agent_test.go
@@ -46,8 +46,8 @@ func TestACPXAgentRunUsesPersistentCodexSession(t *testing.T) {
 	if first.Output != "done from acpx" {
 		t.Fatalf("first Output = %q, want fake acpx message", first.Output)
 	}
-	if first.SessionID != "liza-coder-1" {
-		t.Fatalf("first SessionID = %q, want liza-coder-1", first.SessionID)
+	if first.SessionID != "liza-coder-1-task-acp" {
+		t.Fatalf("first SessionID = %q, want liza-coder-1-task-acp", first.SessionID)
 	}
 	if first.WarmUsage {
 		t.Fatal("first WarmUsage = true, want false")
@@ -68,8 +68,8 @@ func TestACPXAgentRunUsesPersistentCodexSession(t *testing.T) {
 	log := readTextForTest(t, logPath)
 	for _, want := range []string{
 		"ENV_LIZA_AGENT_ID:coder-1",
-		"ARGS:--cwd " + req.ProjectRoot + " codex sessions ensure --name liza-coder-1",
-		"ARGS:--cwd " + req.ProjectRoot + " --format json --approve-all codex prompt -s liza-coder-1 --file -",
+		"ARGS:--cwd " + req.ProjectRoot + " codex sessions ensure --name liza-coder-1-task-acp",
+		"ARGS:--cwd " + req.ProjectRoot + " --format json --approve-all codex prompt -s liza-coder-1-task-acp --file -",
 		"STDIN:implement the requested change",
 	} {
 		if !strings.Contains(log, want) {
--- a/internal/agent/launch_resolver.go
+++ b/internal/agent/launch_resolver.go
@@ -215,7 +215,15 @@ func ResolveLaunchPlan(req LaunchPlanRequest) (LaunchPlan, error) {
 		// every prompt, and auto-repair then respawns into the same poisoned
 		// session (DEV-667). Templates that already place {{taskID}} keep
 		// full control.
-		if scope := acpxSessionTaskScope(req.TaskID); scope != "" && !strings.Contains(sessionTemplate, "{{taskID}}") {
+		templateIncludesTaskID := false
+		for _, match := range templateExprRE.FindAllString(sessionTemplate, -1) {
+			name := strings.TrimSpace(strings.TrimSuffix(strings.TrimPrefix(match, "{{"), "}}"))
+			if name == "taskID" {
+				templateIncludesTaskID = true
+				break
+			}
+		}
+		if scope := acpxSessionTaskScope(req.TaskID); scope != "" && !templateIncludesTaskID {
 			renderedSessionName = renderedSessionName + "-" + scope
 		}
 		vars["sessionName"] = renderedSessionName
--- a/internal/agent/launch_resolver.go
+++ b/internal/agent/launch_resolver.go
@@ -210,6 +210,14 @@ func ResolveLaunchPlan(req LaunchPlanRequest) (LaunchPlan, error) {
 		if err != nil {
 			return LaunchPlan{}, fmt.Errorf("%s acpx session name: %w", toolName, err)
 		}
+		// Scope the session to the task so context cannot accumulate across
+		// tasks: a session that spans tasks grows until the provider rejects
+		// every prompt, and auto-repair then respawns into the same poisoned
+		// session (DEV-667). Templates that already place {{taskID}} keep
+		// full control.
+		if scope := acpxSessionTaskScope(req.TaskID); scope != "" && !strings.Contains(sessionTemplate, "{{taskID}}") {
+			renderedSessionName = renderedSessionName + "-" + scope
+		}
 		vars["sessionName"] = renderedSessionName
 	}
 
@@ -345,6 +353,26 @@ func validatePromptTransport(value string) error {
 	default:
 		return fmt.Errorf("unsupported prompt transport: %s", value)
 	}
+}
+
+// acpxSessionTaskScope converts a task ID into a session-name suffix. Task IDs
+// are engine-generated kebab-case, but sanitize defensively so the name stays
+// safe as an acpx session key.
+func acpxSessionTaskScope(taskID string) string {
+	taskID = strings.ToLower(strings.TrimSpace(taskID))
+	if taskID == "" {
+		return ""
+	}
+	var b strings.Builder
+	for _, r := range taskID {
+		switch {
+		case r >= 'a' && r <= 'z', r >= '0' && r <= '9', r == '-', r == '_', r == '.':
+			b.WriteRune(r)
+		default:
+			b.WriteByte('-')
+		}
+	}
+	return strings.Trim(b.String(), "-")
 }
 
 func launchTemplateVars(req LaunchPlanRequest, toolName string) map[string]string {
--- a/internal/agent/launch_resolver.go
+++ b/internal/agent/launch_resolver.go
@@ -363,24 +363,14 @@ func validatePromptTransport(value string) error {
 	}
 }
 
-// acpxSessionTaskScope converts a task ID into a session-name suffix. Task IDs
-// are engine-generated kebab-case, but sanitize defensively so the name stays
-// safe as an acpx session key.
+// acpxSessionTaskScope preserves the validated task ID so session names stay
+// collision-free across distinct valid task IDs.
 func acpxSessionTaskScope(taskID string) string {
-	taskID = strings.ToLower(strings.TrimSpace(taskID))
+	taskID = strings.TrimSpace(taskID)
 	if taskID == "" {
 		return ""
 	}
-	var b strings.Builder
-	for _, r := range taskID {
-		switch {
-		case r >= 'a' && r <= 'z', r >= '0' && r <= '9', r == '-', r == '_', r == '.':
-			b.WriteRune(r)
-		default:
-			b.WriteByte('-')
-		}
-	}
-	return strings.Trim(b.String(), "-")
+	return taskID
 }
 
 func launchTemplateVars(req LaunchPlanRequest, toolName string) map[string]string {
--- a/internal/agent/launch_resolver_test.go
+++ b/internal/agent/launch_resolver_test.go
@@ -401,6 +401,26 @@ func TestResolveLaunchPlanRespectsExplicitTaskIDTemplate(t *testing.T) {
 	}
 }
 
+func TestResolveLaunchPlanRespectsWhitespaceInExplicitTaskIDTemplate(t *testing.T) {
+	tool := acpxTaskScopeTestTool()
+	tool.ACPXSessionName = "custom-{{ taskID }}-{{agentID}}"
+	config := models.Config{AgentTools: map[string]models.AgentToolConfig{"qwen-acp": tool}}
+
+	plan, err := ResolveLaunchPlan(LaunchPlanRequest{
+		ToolName:      "qwen-acp",
+		AgentID:       "coder-2",
+		TaskID:        "fleet-coding-1",
+		ProjectRoot:   "/repo",
+		RuntimeConfig: config,
+	})
+	if err != nil {
+		t.Fatalf("ResolveLaunchPlan() error = %v", err)
+	}
+	if plan.ACPXSessionName != "custom-fleet-coding-1-coder-2" {
+		t.Fatalf("ACPXSessionName = %q, want custom-fleet-coding-1-coder-2 (no double task suffix)", plan.ACPXSessionName)
+	}
+}
+
 func TestACPXSessionTaskScopeSanitizes(t *testing.T) {
 	for _, tt := range []struct{ in, want string }{
 		{"code-plan-mobile-coding-0", "code-plan-mobile-coding-0"},
--- a/internal/agent/launch_resolver_test.go
+++ b/internal/agent/launch_resolver_test.go
@@ -364,6 +364,42 @@ func TestResolveLaunchPlanScopesACPXSessionToTask(t *testing.T) {
 	}
 }
 
+func TestResolveLaunchPlanScopesDistinctValidTaskIDsToDistinctSessions(t *testing.T) {
+	config := models.Config{AgentTools: map[string]models.AgentToolConfig{"qwen-acp": acpxTaskScopeTestTool()}}
+
+	upperPlan, err := ResolveLaunchPlan(LaunchPlanRequest{
+		ToolName:      "qwen-acp",
+		AgentID:       "coder-1",
+		TaskID:        "Task-1",
+		ProjectRoot:   "/repo",
+		RuntimeConfig: config,
+	})
+	if err != nil {
+		t.Fatalf("ResolveLaunchPlan(upper) error = %v", err)
+	}
+
+	lowerPlan, err := ResolveLaunchPlan(LaunchPlanRequest{
+		ToolName:      "qwen-acp",
+		AgentID:       "coder-1",
+		TaskID:        "task-1",
+		ProjectRoot:   "/repo",
+		RuntimeConfig: config,
+	})
+	if err != nil {
+		t.Fatalf("ResolveLaunchPlan(lower) error = %v", err)
+	}
+
+	if upperPlan.ACPXSessionName != "liza-qwen-coder-1-Task-1" {
+		t.Fatalf("upper session = %q, want liza-qwen-coder-1-Task-1", upperPlan.ACPXSessionName)
+	}
+	if lowerPlan.ACPXSessionName != "liza-qwen-coder-1-task-1" {
+		t.Fatalf("lower session = %q, want liza-qwen-coder-1-task-1", lowerPlan.ACPXSessionName)
+	}
+	if upperPlan.ACPXSessionName == lowerPlan.ACPXSessionName {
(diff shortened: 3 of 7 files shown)


LivioGama:
Follow-up for DEV-667 review findings:

- P1 — templates such as `{{ taskID }}` rendered correctly but were not recognized as explicitly task-scoped, so they received a duplicate task suffix. Fixed in `adb45e6285a593a1e4935e81b993cae096e1ff75` by inspecting template expressions with the same whitespace-tolerant parsing used by rendering; the added regression covers the exact case.
- P1 — an ACPX prompt process error returned as a Go error before the supervisor’s crash tracker, so auto-repair could restart the supervisor into the same task session. Fixed in `4fe76879b5f02d7016e5674f8f1b50a55c095a51`: non-zero agent errors now use the existing crash-loop path, which blocks the task at its configured threshold; the new supervisor regression proves two failures block before process respawn.
- Validation unblock — `.pre-commit-config.yaml` now resolves `gofmt` from `PATH`, replacing the macOS-specific `/usr/local/go/bin/gofmt` path. The full touched-file pre-commit run now passes.

Validation:

- `pre-commit run --files .pre-commit-config.yaml internal/agent/launch_resolver.go internal/agent/launch_resolver_test.go`
- `pre-commit run --files internal/agent/systemctl.go internal/agent/supervisor_test.go`
- `go test ./internal/agent -count=1` — complete package run returned successfully after both corrective commits

tangi-vass:
**Review: Standard (deep pass on the supervisor loop) — Comment**

Independent review at round 1 (no prior `pr-review` chain on this PR). Reviewed head `4fe76879b5f02d7016e5674f8f1b50a55c095a51` against DEV-667. The two findings answered in the follow-up comment reconcile as **RESOLVED** against source, not just the thread: whitespace-tolerant `{{ taskID }}` detection at `internal/agent/launch_resolver.go:204-211` with its regression, and the non-zero-error path at `internal/agent/systemctl.go:243-249`. Every `ACPXAgent.Run` error return sets `ExitCode: 1` (`acpx_agent.go:46,81,111`), so the new branch does cover the failure class the ticket describes. Findings below are new.

Blockers: None
Concerns: 3
Suggestions: 3

---

### `[concern]` specs/architecture/ADR/0085-llm-agent-boundary-and-acp-observability.md:63 — recorded decision now contradicted, no amendment

ADR-0085 states: *"session names are scoped to the Liza agent identity so later turns by the same agent can reuse provider session state."* `launch_resolver.go:199-214` makes them agent+task scoped. The code is the authoritative side here — the ADR's assumption is what produced the incident — but the record still says the opposite, and `specs/architecture/ADR/README.md` shows this repo supersedes decisions explicitly rather than letting them drift.

**Why it matters:** the next change touching ACP session identity reads the ADR, not this PR.
**Likelihood:** certain — the ADR is unchanged at head.
**Closure condition:** ADR-0085 amended (or a superseding ADR added and indexed) recording task-scoped session names and the reason.

### `[concern]` internal/agent/supervisor.go:859-884 — the pre-execution guard shadows the post-execution one

Both guards now consume the same signature. Pre-execution `spinTracker.Track` (line 869) uses `readSuccessProgressSnapshot`; post-execution `successNoProgressTracker.Track` (line 986) uses `successfulTurnProgressSignature(_, _, postSuccessSnapshot)`, which returns `taskSnapshot` unchanged (line 567). Same source, same `spinningTracker.Track` semantics (line 412), same threshold.

In a no-progress loop both counters reach *k* in iteration *k*, but the spin check runs first and `continue`s, so `successNoProgressTracker` never reaches `T+1`. `spinTracker` also counts non-zero-exit iterations that `successNoProgressTracker` skips without resetting, so it can only lead. The `"successful no-progress loop detected"` branch (lines 988-1002) is therefore unreachable for the case it was written for — and this PR renamed its only test (`TestRunSupervisor_OpenCodeSuccessfulNoProgressBlocksAsSpinning`) onto the spin path, so nothing asserts it any more.

**Why it matters:** an alert string, a tracker, and a blocked-reason operators may search for are now dead weight; the next reader will assume both fire.
**Likelihood:** every eligible task — the two signatures are equal by construction, not by coincidence.
**Closure condition:** one guard owns this failure — either retire the shadowed branch and its tracker, or state in code why it stays reachable and cover it.

### `[concern]` .pre-commit-config.yaml:67 — `gofmt` resolution diverges from every sibling Go hook

`entry: gofmt -w -l` relies on the caller's `PATH`. Every other Go hook in the same `local` repo block explicitly prepends it: `goimports` (line 75), `go-vet` (81), `staticcheck` (88), `go-mod-tidy` (102) all `export PATH="/usr/local/go/bin:$PATH"` first, with line 71-72 recording why. The absolute path was the same compensation.

CI is not evidence here: `make lint` runs `go fmt`/`go vet` directly (`Makefile:111-113`); no workflow runs pre-commit.

**Why it matters:** in the sandboxes the neighbouring hooks accommodate, `go-fmt` now fails with `Executable gofmt not found` — trading one environment's breakage for another's.
**Likelihood:** high in any environment without `/usr/local/go/bin` on `PATH`.
**Closure condition:** the hook resolves `gofmt` in both environments, e.g. `entry: bash -c 'export PATH="/usr/local/go/bin:$PATH"; exec gofmt -w -l "$@"' --`.

### `[suggestion]` internal/agent/supervisor.go:233-264 — two signature functions, now one behavior

With `snapshot.Iteration = 0` added, `exit42TaskProgressSignature` and `successfulTurnTaskProgressSignature` differ only in a `json.Marshal`-failure fallback string that cannot be reached for these types. Two names with identical semantics drift apart on the next edit — worth collapsing, or a comment on each saying what is expected to diverge.

### `[suggestion]` internal/agent/systemctl.go:243-249 — the failure reason does not reach the supervisor

`result.Output` for an `acpx prompt` failure is the partial `output.Text`, not the error (`acpx_agent.go:111`); the masked error text is only logged. Downstream, `handleClassifiedProviderCrash(config, currentOutput)` and the eventual crash-loop `BlockedReason` see output without the cause. Appending the masked error to the returned output would let provider classification and the operator-facing reason carry `-32603 Internal error: Bad Request` instead of silence.

### `[suggestion]` session lifecycle — the second half of the ticket's fix 1, and where it is recorded

DEV-667 fix 1 reads *"derive the session name from task identity …; end/rotate the session at task completion"*. The naming half is here; `SessionManager.End()` is still never called, and per-task naming multiplies persisted sessions from one per agent slot to one per agent×task with no reaper. That is accepted as a follow-up in the ticket comment, which is the right call for this PR — but `TECH_DEBT.md` exists in this repo and is where accepted deferrals are recorded with a payback trigger. A line there, or a linked follow-up ticket in the PR body, keeps it from living only in a Jira comment.

### `[question]` internal/agent/systemctl.go:243 — config errors now travel the crash path

`ResolveLaunchPlan` failures return `ExitCode: 1` plus an error (`acpx_agent.go:46`), so a permanently bad `acpx_session_name` template now retries five times and blocks the task, then does the same to the next task claimed, instead of failing the supervisor loudly. Deliberate, or worth short-circuiting a plan-resolution error before the crash tracker?

### `[appraisal]`

`TestExit42TaskProgressSignatureIgnoresClaimIteration` asserts the mechanism that failed — a claim-only `Iteration` bump must not change the signature, real status progress must — rather than the shape of the fix. That test would have caught the original bug.

---

**Overall:** the two root causes are addressed at the right boundaries, and the loop breakers go from never firing to firing on real evidence. What remains is bookkeeping around the change, not the change itself.
**Approach:** sound. The alternative weighed was bounding the session by token/turn budget instead of by task identity — it caps the symptom but leaves cross-task history, and DEV-667 already lists it as defence-in-depth rather than the fix.
**Absence:** baseline is DEV-667's "Fix (this ticket)" list. Item 2 complete; item 1 half-complete (see the session-lifecycle suggestion). The two changes beyond that list — the worktree-aware spin signature and the `systemctl.go` error path — are declared and justified.
**Net value:** *As submitted* — replaces an unbounded cross-task ACP session and three structurally disabled loop breakers with per-task sessions and counters that actually fire, retaining one shadowed guard, an unamended ADR, and a pre-commit hook that now depends on `PATH`. *With findings resolved* — the same benefit with the architecture record and guard ownership consistent, at no added runtime complexity.
**Blast Radius:** High — production agent runtime, on paths that can move a task to BLOCKED.
**Confidence:** high on the Go changes; medium on the pre-commit hook, which no CI job exercises.
**Sources:** PR diff at head; DEV-667 and its comment; `launch_resolver.go`, `supervisor.go`, `systemctl.go`, `acpx_agent.go`, `cli_agent.go` at head; `progress_watchdog.go`, `models/task.go`, `models/config.go`, `.pre-commit-config.yaml`, `Makefile`, `.github/workflows/ci.yml`, ADR-0085, `specs/architecture/acp-vs-cli.md` on `main` (unchanged by this PR); check runs `Test (ubuntu-latest)` and `Test (macos-latest)` — both success at this head; `main` ruleset requires 1 approving review and no status checks.
**Next step:** address or explicitly accept the three concerns; nothing here needs another full pass.

tangi-vass:
**Review: Deep — Request Changes**

Independent round-1 review of head `4fe76879b5f02d7016e5674f8f1b50a55c095a51` against DEV-667.

A separate independent review comment appeared on this same head while this pass was underway. Because the head has not changed, its concerns and suggestions remain unaddressed; I am not duplicating them here. This review adds one distinct correctness blocker.

Blockers: 1
Concerns: 3 existing, not repeated
Suggestions: 3 existing, not repeated

---

### `[blocker]` internal/agent/launch_resolver.go:348 — distinct valid task IDs can share an ACP session

`acpxSessionTaskScope` lowercases the task ID and trims surrounding hyphens. That mapping is not one-to-one: `Task-1` and `task-1` both become `task-1`; `task` and `task-` both become `task`.

All four values pass `paths.ValidateTaskID`, `ops.AddTask` preserves the supplied ID, and state uniqueness compares the original IDs exactly. With the default ACP session template, the same agent therefore resolves either task in each pair to the same session key.

**Why it matters:** DEV-667 requires fresh context per task. These collisions retain the cross-task context leakage and poisoned-session reuse that this PR is intended to eliminate.

**Likelihood:** reachable through accepted task input; unusual naming is enough, with no malformed state or configuration required.

**Closure condition:** session-scope derivation is collision-free across every valid task ID, with regression tests showing that case-distinct and trailing-hyphen task IDs produce distinct session names. Preserving the validated ID or using a collision-safe encoding are both acceptable implementations.

---

**Overall:** the root approach is sound, and the supervisor-loop changes repair the failure mode at appropriate boundaries. The task suffix must preserve task identity rather than a lossy normalization of it.

**Approach:** task-scoped sessions are the right first-line fix. A session-budget alternative would bound growth but would not prevent cross-task context sharing.

**Absence:** DEV-667 is the baseline. Coverage is missing for the invariant that every pair of distinct valid task IDs produces distinct default session names.

**Sweep:** not applicable — independent round 1.

**Net value:** *As submitted* — common task IDs receive isolated sessions and the loop breakers become effective, but accepted task IDs can still recreate the incident’s cross-task session sharing. *With the blocker resolved* — the PR provides task-level context isolation and functioning no-progress detection with minimal additional complexity.

**Blast Radius:** High — production agent runtime and task-blocking behavior.

**Confidence:** high — complete diff and exact-head source reviewed; task validation and creation paths traced; both exact-head CI jobs succeeded.

**Sources:** PR description, commits, complete diff, comments, and DEV-667; `launch_resolver.go`, `paths.go`, `add_tasks.go`, `statevalidate/validate.go`, supervisor/agent execution paths and tests; relevant invariants and ADRs; GitHub ruleset and exact-head CI results.

**Next step:** make task-session derivation collision-free, add the negative regressions, and address or explicitly accept the concerns in the earlier review comment.

LivioGama:
Addressed the review comments on this PR.

- ACPX session names now preserve the validated task ID verbatim, so distinct valid task IDs do not collide.
- Added regressions for case-distinct task IDs and trailing-hyphen task IDs.
- Removed the dead post-success no-progress branch and aligned the ADR text with task-scoped ACPX sessions.
- Made the `gofmt` pre-commit hook resolve via `PATH` so the file-scoped hook run passes here.

Verified with:
- `pre-commit run --files internal/agent/launch_resolver.go internal/agent/launch_resolver_test.go internal/agent/supervisor.go`
- `go test ./internal/agent -count=1`

tangi-vass (dismissed):
**Review: Standard (round 2 — reconciliation) — Approve**

Reviewed head `ef564df595f5e3d065cc3d892c8ea54961771ef7` against DEV-667.

This round reconciles both round-1 comments on head `4fe7687` — chains `pr103-20260810-171400Z` and `pr103-20260810-180900Z-collision` — which merge into this chain. Depth is reconciliation, not re-derivation: the corrective commit `ef564df` is bounded, and the full change set was swept for P0-P2.

**Round 2 — blocking findings 4→0, files 8→8.**

---

## Reconciliation

| Round-1 finding | Status |
|---|---|
| `[blocker]` distinct valid task IDs share an ACP session | **RESOLVED** |
| `[concern]` ADR-0085 contradicted by the code | **RESOLVED** |
| `[concern]` post-success no-progress guard shadowed by the spin guard | **RESOLVED** |
| `[concern]` `gofmt` resolution diverges from sibling Go hooks | **RESOLVED** |
| `[suggestion]` two signature functions, one behavior | STILL PRESENT → follow-up |
| `[suggestion]` ACPX failure reason does not reach the supervisor | STILL PRESENT → follow-up |
| `[suggestion]` session lifecycle / `End()` recorded as debt | STILL PRESENT → follow-up |
| `[question]` config errors travel the crash path | unanswered → non-blocking |

Verified against exact-head source, not thread state:

- **Collision.** `acpxSessionTaskScope` (`internal/agent/launch_resolver.go:346`) now returns the trimmed task ID verbatim. `paths.ValidateTaskID` (`internal/paths/paths.go:315-331`) admits only `[a-zA-Z0-9._-]` after a leading alphanumeric and rejects `..`, so the dropped sanitizer was covering inputs that cannot reach state, and identity-preserving derivation is injective by construction. `TestResolveLaunchPlanScopesDistinctValidTaskIDsToDistinctSessions` pins `Task-1` vs `task-1`; `TestACPXSessionTaskScopePreservesIdentity` pins the trailing-hyphen case. Both negative regressions asked for are present.
- **Shadowed guard.** The `successNoProgressTracker`, its snapshot closure, and the `"successful no-progress loop detected"` branch are gone. The inlining at `supervisor.go:851` is behaviour-preserving: `readSuccessfulTurnProgressSnapshot` returns `("", false, err)` on its only error path (`progress_watchdog.go:156-160`, unchanged by this PR), which is exactly what the removed closure produced. One guard now owns the failure.
- **ADR.** `ADR-0085:64` now records task-scoped session identity and the collision rationale.
- **`gofmt`.** `.pre-commit-config.yaml:67` now matches the `goimports` sibling pattern exactly. Ran the entry as written against a deliberately misformatted file: filenames arrive as `"$@"`, the file is rewritten and listed, exit 0. Resolves in both environments.

**Scope.** `ef564df` touches `.pre-commit-config.yaml`, `launch_resolver.go` (+test), `supervisor.go`, and ADR-0085 — every one named in a round-1 finding, plus that file's tests. No `[overreach]`; the session-scope fix is strictly smaller than what it replaced.

---

## Remaining — none blocking

### `[suggestion]` internal/agent/supervisor.go:567 — `successfulTurnProgressSignature` is now production-dead

Removing the post-success branch removed its only production call site. The function survives at head as `func(_, _, taskSnapshot string) string { return taskSnapshot }`, kept alive solely by `TestSuccessfulTurnProgressSignatureIgnoresOpenCodeOutputVariation` (`supervisor_test.go:1506`) — a test asserting that a CLI name and agent output do not affect a signature that nothing computes any more. Note this is distinct from `successfulTurnTaskProgressSignature` (`supervisor.go:253`), which is still live via `progress_watchdog.go:176`. Leftover of the fix rather than a defect; worth removing with its test in the same follow-up as the two-signature-functions suggestion.

### `[nit]` specs/architecture/ADR/0085-llm-agent-boundary-and-acp-observability.md:64 — the agent half of the scoping dropped out

"scoped to the Liza task identity" — session names are scoped to agent **and** task (`liza-{{agentID}}` + task suffix), and taskless runs stay agent-scoped. The agent component still does work: a coder and a reviewer touching the same task ID resolve to different sessions. Worth restoring the conjunction so the next reader does not infer task-only keying.

### `[suggestion]` collision guarantee ends at the name

The fix makes derivation injective over *session names*. Whether two names differing only in case resolve to two provider sessions depends on `acpx` session storage, which is outside this repo and which I did not verify. Likelihood low — engine-generated task IDs are lowercase kebab-case, so case-distinct pairs are a theoretical input, and the round-1 blocker was about admissible rather than observed IDs. Recording where the guarantee stops, not asking for a change.

### `[appraisal]`

`TestResolveLaunchPlanScopesDistinctValidTaskIDsToDistinctSessions` asserts the invariant rather than the implementation — it fails for *any* lossy derivation, not just the lowercasing one that was there. That is the version of the test that keeps holding after the next refactor.

---

**Overall:** every blocking finding from round 1 closes against source, and the two that mattered most closed by removing code rather than adding it. What remains is one orphaned function, one ADR sentence, and three follow-ups already accepted in DEV-667.

**Approach:** unchanged and sound. Per-task session identity plus claim-churn-free progress signatures address both root causes at their own boundaries.

**Absence:** DEV-667's "Fix (this ticket)" list remains the baseline. Item 2 complete; item 1's naming half complete, its `End()`/rotation half still deferred without a `TECH_DEBT.md` entry or a linked follow-up ticket — accepted for this PR, still the one piece of bookkeeping that lives only in a Jira comment.

**Sweep:** full change set re-inspected for P0-P2. No new blocking issue introduced by the corrective commit.

**Net value:** replaces an unbounded cross-task ACP session and three structurally disabled loop breakers with per-task sessions and counters that fire on real evidence — now without the lossy task-ID mapping, the unreachable second guard, the stale architecture record, or the `PATH`-dependent hook.

**Blast Radius:** High — production agent runtime, on paths that can move a task to BLOCKED.

**Confidence:** high. Complete diff and exact-head source read by SHA; the removed guard's equivalence traced through `progress_watchdog.go`; the hook entry executed as written. No local exact-head worktree was created: both required-path CI jobs (`Test (ubuntu-latest)`, `Test (macos-latest)`) are green at this head and cover the Go tests, and the one thing CI does not exercise — the pre-commit hook — was run directly here instead.

**Sources:** PR description, four commits, complete diff, all issue comments, and DEV-667 with its comment; `launch_resolver.go`, `supervisor.go`, `paths.go`, `.pre-commit-config.yaml`, ADR-0085 at head `ef564df`; `progress_watchdog.go` and `TECH_DEBT.md` (unchanged by this PR); `protect-main` ruleset — 1 approving review, no required status checks; check runs at this head both successful.

**Next step:** merge. The three follow-up suggestions and the ADR nit do not need another pass.

tangi-vass (approved):

Issue #104, opened 2026-08-06
Revisit native Windows support?


jgicquel:
I've been using Liza on a Windows machine where WSL2 isn't available, so I spent some time getting the suite to run natively on windows/amd64. It works well enough now that I'd like to offer the work back, if that's of interest.

I understand why v0.8.0 went the other way (788a9f4): the suite didn't pass there, nothing covered the platform in CI, and WSL2 makes the problem go away cheaply. That reasoning holds in most situations. The one it doesn't cover is a managed machine where WSL2 simply isn't an option, which is where I ended up.

The more useful part of this may not be the portability work itself. Getting there meant instrumenting a few code paths that Windows exercises differently, and a couple of things surfaced that I don't think are specific to Windows. I could be missing context on some of them, so I've written down what I observed rather than what I concluded.

## What I ran into

**1. Session initialization doesn't get past the gate on Windows.**
`enforce-init.sh` treats the backslash as a shell metacharacter, and `is_session_init_doc_path` compares against `$project_dir`, which `git rev-parse --show-toplevel` reports forward-slashed even on Windows. Between the two, a Bash command carrying a native path doesn't make it through. I found this by instrumenting the hook rather than by reading it, so if the guard is shaped that way deliberately I'd be glad to hear it.

**2. `session-context.sh` mixes separators.**
`$project_dir` falls back to the agent-supplied `cwd`, which is backslashed, while everything appended to it is joined with `/`. Agents end up reading lines like `Stacklit index: C:\Users\me\proj/stacklit.json`, and the emitted commands carry the same mixed path inside quotes.

**3. The one I'd most like a second opinion on: managed git hooks.**
`installManagedHook` removes the existing hook before attempting `os.Symlink`. When the symlink can't be created, the wrapper fallback finds nothing there, reinstalls from scratch, and reports `installed` rather than `verified`/`updated` — on every run. Separately, the wrapper execs the dispatcher under the dispatcher's own name, so `basename "$0"` is never the hook name and the `post-checkout` / `$3 = 0` short-circuit doesn't fire.

Windows hits this every time, which is why I noticed it, but as far as I can tell any environment where `os.Symlink` fails would take the same path. That's the part I think might matter beyond my setup.

**4. A smaller one, for contributors rather than users.**
[...]

PR #105, opened 2026-08-07
fix: keep managed git hooks intact and derive the build version from git

jgicquel:
Two independent defects, both found while getting the suite to run on Windows, neither specific to it. Sending them separately from the Windows work so they can be judged on their own.

Follows up on #104.

## 1. A managed git hook is destroyed before its replacement exists

`installManagedHook` removes the existing hook before calling `os.Symlink`, because `os.Symlink` cannot overwrite. When the symlink then fails, the wrapper fallback finds nothing where the hook used to be, reinstalls from scratch, and reports `installed` — every run.

Two consequences:

- A hook the user may have been running for months is deleted and recreated rather than verified.
- `verified` and `updated` never appear on that path, so the report is wrong about what happened.

The link is now created beside the hook and renamed over it. A failure to create it leaves the existing hook untouched, and the fallback sees the real previous state.

**The wrapper also loses the hook name.** It execs the dispatcher under the dispatcher's own name, so `basename "$0"` is never the hook that ran, and the dispatcher's post-checkout short-circuit — skip indexing when `$3` is `0`, a file checkout rather than a branch switch — cannot fire. Every `git checkout -- <file>` triggers a full re-index. The wrapper now passes the name in `PAIRING_INDEX_HOOK_NAME`; the dispatcher prefers it and falls back to `basename $0` for hooks installed as symlinks.

Existing installations migrate on the next run: the dispatcher content differs from the rendered one, carries the managed marker, and is rewritten as `updated`.

This path is reached wherever `os.Symlink` fails consistently. On Windows that is the default, which is how I ran into it, but nothing in the fix is Windows-specific.

## 2. Every source build claims version 0.2.0

`VERSION?=0.2.0` is pinned in the Makefile, so `make build` has reported 0.2.0 since long after 0.8.0 shipped.

That is not only cosmetic. The updater compares the running version against the latest release, so a binary reporting 0.2.0 believes it is several minor versions behind and offers to replace itself with the published release. The check is opt-in, so nothing fires by default, but the behaviour is armed for anyone who enables it.

`VERSION` now comes from git. Only an exact, clean tag — HEAD untouched since that tag, no local changes — reports the real version. Everywhere else (commits past the tag, a dirty tree, no tags, no git at all) reports the non-semver `dev-<short-sha>` (`dev-unknown` without git). An earlier version of this fix used `git describe --tags --always --dirty` directly, which reports a valid-but-prerelease semver value like `v0.8.0-3-gabc123` for any commit past a tag — `semver.Compare` sorts a prerelease below the release it precedes, so `lookupCandidate` judged that as *older* than the last release and would still offer to replace a newer dev build with an older one, which is the exact regression this fix exists to prevent. The exact-tag-only form closes that. Release builds are unaffected: GoReleaser passes its own version from the tag and never reads this variable.

## Change map

| File | What |
|------|------|
| `internal/pairingindex/hooks.go` | Stage the symlink and rename it over the hook; pass the hook name to the dispatcher |
| `Makefile` | Derive `VERSION` from git |

## Validation

- `go build ./...` and `go vet ./internal/pairingindex/` clean.
- `internal/pairingindex` on windows/amd64: 19 failures on `main` alone, the same 19 with these two commits, compared by test name. No regression. Those 19 are the pre-existing Windows symlink and PATHEXT problems, which are not addressed here.
- I could not exercise the Unix symlink path locally — this machine has neither Developer Mode nor the privilege — so CI is the check that matters for the happy path.

## Risks

The hook change alters the order of operations on a path that is exercised on every `init`. On a host where `os.Symlink` succeeds, the observable outcome is the same symlink as before, created through a rename over a per-invocation-unique staging name (via `os.CreateTemp`) rather than a fixed `<hook>.tmp` — an earlier version of this fix used the fixed name, which could delete an unrelated file that happened to exist there. The rename protects the existing hook if creating the new symlink fails; `os.Rename` is not documented as atomic on non-Unix platforms even within one directory, so this is not a portable atomic-replacement guarantee — only that a failed replacement leaves the previous hook in place.

## Not in scope

Native Windows support itself. That is a much larger change — portability, test equivalents, CI, packaging — and I would rather send it once it is green than drag these two fixes behind it.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -18,7 +18,12 @@ BRAND_CHECKSUM_BASE_URL?=$(BRAND_RELEASE_BASE_URL)
 BINARY_NAME?=$(BRAND_BINARY_NAME)
 
 # Build variables
-VERSION?=0.2.0
+# Derived from git rather than pinned, because a pinned default goes stale and
+# then lies: the binary claimed 0.2.0 long after 0.8.0 shipped. The fallback is
+# deliberately not semver — a hand-built binary is not a release, and the
+# updater only considers a valid semver current version, so a source build is
+# left alone instead of being offered a release to overwrite itself with.
+VERSION?=$(shell git describe --tags --always --dirty 2>/dev/null || echo "dev")
 GIT_COMMIT?=$(shell git rev-parse --short HEAD 2>/dev/null || echo "unknown")
 BUILD_DATE?=$(shell date -u +"%Y-%m-%dT%H:%M:%SZ")
 LDFLAGS=-ldflags "-X 'github.com/liza-mas/liza/internal/embedded.Version=$(VERSION)' \
--- a/Makefile
+++ b/Makefile
@@ -19,11 +19,15 @@ BINARY_NAME?=$(BRAND_BINARY_NAME)
 
 # Build variables
 # Derived from git rather than pinned, because a pinned default goes stale and
-# then lies: the binary claimed 0.2.0 long after 0.8.0 shipped. The fallback is
-# deliberately not semver — a hand-built binary is not a release, and the
-# updater only considers a valid semver current version, so a source build is
-# left alone instead of being offered a release to overwrite itself with.
-VERSION?=$(shell git describe --tags --always --dirty 2>/dev/null || echo "dev")
+# then lies: the binary claimed 0.2.0 long after 0.8.0 shipped. Only an exact,
+# clean tag produces the semver value — anything else (commits past the tag, a
+# dirty tree, no tags, no git) is the non-semver "dev-<sha>". A prerelease-style
+# value like "v0.8.0-3-gabc123" would still be valid semver but would sort
+# *below* the v0.8.0 release, so the updater would offer to replace a newer
+# dev build with an older one. Non-semver keeps the updater's "valid semver
+# current version required" gate closed, so a source build is left alone
+# instead of being offered a release to overwrite itself with.
+VERSION?=$(shell tag=$$(git describe --tags --exact-match 2>/dev/null); if [ -n "$$tag" ] && [ -z "$$(git status --porcelain 2>/dev/null)" ]; then printf '%s' "$$tag"; else printf 'dev-%s' "$$(git rev-parse --short HEAD 2>/dev/null || echo unknown)"; fi)
 GIT_COMMIT?=$(shell git rev-parse --short HEAD 2>/dev/null || echo "unknown")
 BUILD_DATE?=$(shell date -u +"%Y-%m-%dT%H:%M:%SZ")
 LDFLAGS=-ldflags "-X 'github.com/liza-mas/liza/internal/embedded.Version=$(VERSION)' \
--- a/internal/pairingindex/hooks.go
+++ b/internal/pairingindex/hooks.go
@@ -955,8 +955,17 @@ func installManagedHook(hookPath, hook string) (HookAction, error) {
 	// or an elevated shell, and filesystems that do not support them. The
 	// wrapper fallback then sees no file, reinstalls from scratch and reports
 	// "installed" on every run instead of "verified" or "updated".
-	staged := hookPath + ".tmp"
-	if err := os.Remove(staged); err != nil && !os.IsNotExist(err) {
+	//
+	// The staging name is unique per invocation (rather than a fixed
+	// "<hook>.tmp") so this never deletes a file it did not create itself.
+	dir := filepath.Dir(hookPath)
+	stagingFile, err := os.CreateTemp(dir, filepath.Base(hookPath)+".tmp-*")
+	if err != nil {
+		return "", fmt.Errorf("stage %s hook: %w", hook, err)
+	}
+	staged := stagingFile.Name()
+	stagingFile.Close()
+	if err := os.Remove(staged); err != nil {
 		return "", fmt.Errorf("clear staged %s hook: %w", hook, err)
 	}
 	if err := os.Symlink(hookDispatcherName(), staged); err != nil {
@@ -1022,6 +1031,7 @@ if [ -z "$repo_root" ]; then
 fi
 
 hook_name="${%s:-$(basename "$0")}"
+unset %s
 if [ "$hook_name" = "post-checkout" ] && [ "${3:-}" = "0" ]; then
 	exit 0
 fi
@@ -1040,7 +1050,7 @@ fi
 
 cd "$repo_root"
 "$script"
-`, ManagedHookMarker, hookNameEnvVar, scriptName())
+`, ManagedHookMarker, hookNameEnvVar, hookNameEnvVar, scriptName())
 }
 
 func looksLikeLegacyHookDispatcher(content string) bool {
--- a/internal/pairingindex/hooks.go
+++ b/internal/pairingindex/hooks.go
@@ -23,6 +23,15 @@ const ManagedIndexScriptMarker = "# PAIRING-INDEX-SCRIPT: managed"
 
 const legacyManagedIndexScriptMarker = "# LIZA-PAIRING-INDEX-SCRIPT: managed"
 
+// hookNameEnvVar carries the git hook name from the wrapper to the dispatcher.
+//
+// When the hook is a symlink to the dispatcher, the dispatcher reads the name
+// from $0. The wrapper installed where symlinks are unavailable execs the
+// dispatcher by its own name, so $0 says "liza-index-hook.sh" and the
+// post-checkout file-checkout short-circuit would never fire. The wrapper
+// therefore states the name explicitly.
+const hookNameEnvVar = "PAIRING_INDEX_HOOK_NAME"
+
 const stacklitArtifactName = "stacklit.json"
 const stacklitInsightsArtifactName = "stacklit-insights.json"
 const stacklitArchitectureArtifactName = "stacklit-architecture.json"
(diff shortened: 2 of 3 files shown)


tangi-vass:
**Review: Standard — Comment** (head `5941c5d5`)

Blockers: None · Concerns: 2 · Suggestions: 1 · Nits: 2

Both defects are real and both fixes are the minimum that makes them false. The hook fix I verified by execution. The version fix does not fully deliver what §2 claims, and neither change carries a test.

**Blast radius:** Medium — hook install runs on every `init`; the version change reaches the updater. **Confidence:** high on the hook path (rendered and ran the generated scripts, ran the package suite against this head on Linux), high on the version path (measured against `internal/updater` helpers).

---

### Validation status

CI has not run for this head. Run `31157217017` concluded `action_required` — fork PRs need maintainer approval, so `make lint`, `make test`, `make test-e2e` and `make build` never executed. The PR notes "CI is the check that matters for the happy path"; that check is still outstanding and a maintainer has to release it.

Filling part of that gap locally: `internal/pairingindex` at this head on linux/amd64 — **28 passed**, `go vet` clean. That includes `TestInstallLifecycleHooksRefreshesStaleManagedHook` and `TestInstallLifecycleHooksIsIdempotentForManagedHooks`, which exercise the new stage-and-rename path where `os.Symlink` succeeds. The Unix happy path holds.

---

### `[concern]` `Makefile:26` — the version change does not disarm the updater on ordinary source builds, and the comment at `:21-25` says it does

Measured on a current checkout: `git describe --tags --always --dirty` → `v0.8.0-132-gc7e99f19-dirty`. Against `internal/updater`:

```
raw="v0.8.0-132-gc7e99f19-dirty"  normalized=…  valid=true   newerThan_v0.8.0=true
raw="v0.8.0-dirty"                normalized=…  valid=true   newerThan_v0.8.0=true
raw="c7e99f19"                    normalized=…  valid=false  newerThan_v0.8.0=false
raw="dev"                         normalized=""  valid=false  newerThan_v0.8.0=false
```

`v0.8.0-132-g…` parses as a *prerelease* of 0.8.0, so `semver.Compare` sorts it **below** the v0.8.0 release. A `make build` binary on any commit that is not an exact clean tag is therefore still offered the published release — and now that is a genuine content downgrade (newer dev code replaced by an older release), not only a version-number lie. A dirty tagged checkout (`v0.8.0-dirty`) behaves the same way.

The three cases enumerated in §2 — tagged, untagged, no git — omit the one `make build` actually produces on a development checkout, which is the common case.

**Why it matters:** the exposure is no worse than today (opt-in, interactive, confirmed), so this is not blocking. But §2 is written as closing the hole, and the in-file comment tells the next reader it is closed. That is the part that will mislead.

**Closure condition:** either a non-exact-tag build produces a version `lookupCandidate` will not act on, or the PR body and the `# …left alone instead of being offered a release to overwrite itself with` comment are corrected to state that the exposure persists for non-tag builds. The code is authoritative here — the description is what is wrong, unless closing the hole was the intent.

**Suggestion** (advisory — both verified against the same helpers): emit semver build metadata for non-tag builds, `v0.8.0+132.gc7e99f19`, which is valid semver and compares *equal* to the release so no candidate is found; or make the non-release form plainly non-semver, e.g. `dev-c7e99f19`.

### `[concern]` `internal/pairingindex/hooks.go:1001-1043` — the generated hook contract changed with no test

The wrapper now injects `PAIRING_INDEX_HOOK_NAME` and the dispatcher consumes it. Both are `fmt.Sprintf` shell templates, and nothing in `hooks_test.go` asserts either the rendered content or the post-checkout short-circuit reached through a wrapper.

Rendering both templates and running them confirms the fix works today:

```
$3=0 → (no output)        # file checkout, correctly skipped
$3=1 → INDEX SCRIPT RAN   # branch switch
```

`sh -n` is clean on both. So this is not a correctness finding — it is that the behaviour §1 exists to restore has no guard, and the next edit to those format strings can silently break the short-circuit again on exactly the platform where it cannot be reproduced. `managedHookContent` and `managedHookDispatcherContent` are pure functions; executing the rendered wrapper under `sh` with `a b 0` and `a b 1` is enough.

**Closure condition:** a test asserting the wrapper renders the hook name and that the dispatcher honours it for the post-checkout `$3=0` skip.

### `[suggestion]` `internal/pairingindex/hooks.go:958` — fixed `.tmp` staging name

Two installs running concurrently in one repository race on `hookPath + ".tmp"`: one's pre-clean `os.Remove` can delete the other's staged link between `os.Symlink` and `os.Rename`, failing the second with `replace <hook> hook`. No corruption — the original hook survives, which is the property this PR adds — and concurrent `init` on one repository is unlikely. A unique suffix would remove both the race and the need for the pre-clean step.

### `[nit]` the hook-name variable propagates past the dispatcher

A prefix assignment exports the variable to the dispatcher *and all its descendants*, so a managed hook triggered beneath the index script would read the outer hook's name. Only `post-checkout` keys on the name, and none of `post-commit`, `post-merge`, `post-rewrite` receives a third positional argument, so there is no effect today. An `unset` after the read would bound it.

### `[nit]` `v`-prefix and CI divergence

GoReleaser injects `{{.Version}}` (`0.8.0`); `make build` on that same tag now injects `v0.8.0`. `normalizeVersion` tolerates both and the value only reaches `liza version` and contract frontmatter, so nothing breaks — but two builds of one commit now report differently. Separately, `actions/checkout@v4` fetches no tags, so `make build` in CI will report a bare SHA rather than a version.

### `[appraisal]`

Splitting these out of the Windows work was the right call — they are judgeable on their own, and the write-up makes the causal chain from `os.Symlink` cannot-overwrite → destroyed hook → permanent `installed` easy to follow. The stage-and-rename reorder and the single environment variable are each the smallest change that makes the defect false.

---

**Next step:** correct §2's claim (or close the hole), add the template test, and get a maintainer to release the CI run.

tangi-vass:
**Review: Standard — Request Changes** (head `5941c5d5`)

Blockers: 1 · Concerns: 2 · Suggestions: 1

**Overall:** The two fixes are directionally valuable, but the replacement path introduces a concrete user-data-loss case. The two concerns from the [previous review](https://github.com/liza-mas/liza/pull/105#issuecomment-5240558468) also remain unanswered.

**Approach:** Staging before replacement is the right shape, but the staging path must be invocation-owned rather than a fixed adjacent filename.

**Absence:** The wrapper-specific post-checkout behavior still lacks the test identified previously.

**Net value:** As submitted, the hook-preservation benefit is offset by deletion of an unrelated file at the staging pathname. With the blocker and prior concerns resolved, the PR removes two real defects with little retained complexity.

**Blast radius:** Medium — managed hooks are installed during every `init`, and source-build versions feed the updater.

**Confidence:** High — complete two-file diff, exact-head source and tests, linked issue, prior discussion, branch rules, and local validation inspected.

### Prior-review reconciliation

No commits or author responses followed the previous review; both concerns are **STILL PRESENT**. The earlier fixed-`.tmp` suggestion is superseded by the blocker below because the same line can delete user-owned data, not merely race with another installer.

### `[blocker]` `internal/pairingindex/hooks.go:958-960` — replacement deletes an unowned adjacent file

`staged := hookPath + ".tmp"` selects a predictable path, then `os.Remove(staged)` deletes whatever is there without checking that Liza created or owns it. For example, refreshing a managed `post-merge` wrapper deletes an unrelated `.git/hooks/post-merge.tmp` backup before attempting the symlink.

**Why it matters:** this is irreversible deletion of user-owned data during `init`, contrary to the PR’s purpose of preserving the existing hook until its replacement exists.

**Likelihood:** reached whenever the current hook is managed but not already the expected dispatcher symlink and the adjacent `.tmp` pathname exists.

**Closure condition:** use an invocation-unique staging pathname, clean up only that invocation’s path, and add a test proving a pre-existing `<hook>.tmp` file survives while the managed hook is refreshed.

### `[suggestion]` PR Risks / `internal/pairingindex/hooks.go:965` — narrow the portable atomicity claim

The PR says a same-directory rename is atomic. Go’s `os.Rename` documentation explicitly says that even within one directory it is not atomic on non-Unix platforms. The reorder still protects the old hook when symlink creation fails, but the description should not claim a portable atomic replacement guarantee.

### Validation and merge gates

- Exact head: `5941c5d5b07791384362fe68484fe332bb8a1c4f`
- `go test -race ./internal/pairingindex ./internal/updater`: 138 passed across two packages
- `go vet ./internal/pairingindex ./internal/updater`: passed
- `git diff --check`: passed
- Required `license/cla` status remains pending
- CI run `31157217017` remains `action_required`; its build and test jobs did not execute

**Next step:** replace the fixed staging path, address the two prior concerns, and release CI after the CLA status clears.

tangi-vass (approved):
**Review: Standard — Approve** (head `b99750fe`)

Round 1 (independent re-review) — prior blockers/concerns 3→0, files 2→3.

All prior actionable findings are resolved:

- **RESOLVED:** replacement staging is invocation-owned via `os.CreateTemp`, and the regression test proves an unrelated `<hook>.tmp` survives.
- **RESOLVED:** non-exact-tag builds now report non-semver `dev-<sha>`; the updater’s semver gate stays closed.
- **RESOLVED:** the wrapper/dispatcher path is exercised for both post-checkout file and branch checkouts.
- **RESOLVED:** the dispatcher unsets the hook-name variable before invoking the index script.
- **RESOLVED:** the PR no longer claims portable rename atomicity.
- **ACCEPTED (non-blocking):** exact-tag source builds may retain a `v` prefix while GoReleaser does not; updater normalization handles both, and CI’s `dev-<sha>` is intentional.

No new P0–P2 defects found. The `CreateTemp` → remove → symlink → rename sequence is the minimum stdlib shape that preserves the existing managed hook when symlink creation fails.

**Absence:** both changed behaviors now have regression coverage; no persisted documentation describes the old source-version default.

**Net value:** As submitted, positive — it prevents managed-hook loss/reinstallation and source-build downgrade offers with modest localized complexity. With findings resolved, unchanged — no findings remain.

**Validation:** exact-head diff and source inspected; `git diff --check` passed; repository CI passed `make lint`, `make test`, `make test-e2e`, and `make build` on Ubuntu and macOS; required `license/cla` passed. All applicable pre-commit hooks passed, with staticcheck run under `GOFLAGS=-buildvcs=false` solely because Go 1.26 misidentified the disposable `/tmp` worktree’s VCS root.

PR #106, opened 2026-08-08
test(agent): stop a leaked agent goroutine from racing with brand values

jgicquel:
## Summary

Two data races in `internal/agent` under `-race`, reported against a test that is not at fault. `TestACPXAgentRunStreamsOutputToLogsEventsAndProgress` abandons its agent goroutine when an assertion fails; the abandoned agent keeps reading process-wide brand values while `TestEmitCheckpointSummary_UsesBrandedProjectPathsInPrompt` mutates and restores them.

Two commits: the leak, then the helper that makes the leak dangerous.

## Problem

The streaming test starts `ACPXAgent.Run` in a goroutine and waits for the first progress event, for the run to finish, or for a timeout. The first two paths are fine. The timeout path calls `t.Fatal`, which ends the test immediately — with the goroutine still running and its context never cancelled.

That goroutine calls `agentProcessEnv` → `brand.EnvName` → `brand.RuntimeValues`, which reads `brand.BinaryName` and `brand.ProjectDirName`. `withAgentBrandValues` in `checkpoint_summary_test.go` writes those same package-level variables and restores them from a `t.Cleanup`. When the two overlap:

```
WARNING: DATA RACE
Write at 0x000140ac0af0 by goroutine 363:
  internal/agent.withAgentBrandValues.func1()
      internal/agent/checkpoint_summary_test.go:21
  testing.(*common).runCleanup()

Previous read at 0x000140ac0af0 by goroutine 358:
  internal/brand.RuntimeValues()
      internal/brand/brand.go:71
  internal/agent.agentProcessEnv()
      internal/agent/cli_agent.go:492
  internal/agent.(*ACPXAgent).runACPX()
      internal/agent/acpx_agent.go:261
  internal/agent.(*ACPXAgent).Run()
      internal/agent/acpx_agent.go:67
```

One race per variable, plus a failure of the checkpoint-summary test alongside them. Both tests pass in isolation with no race — the report names the victim, not the cause, which is what makes this one awkward to chase.

The timeout is the trigger, and two seconds is short for a path that starts a real process and waits for it to stream. On an idle machine it never fires; under a full parallel suite it does.

## Approach

**1. Make the abandoned goroutine impossible rather than unlikely.** The run context becomes cancellable, and a `t.Cleanup` cancels it and waits for the goroutine to return, so no assertion can leave an agent behind. `runACPX` and `runACPXPrompt` already use `exec.CommandContext`, so cancelling stops the process and the wait cannot hang. The wait for first progress goes from 2s to 30s — a ceiling on a real process starting up, not a sleep: the test still returns as soon as the event arrives.

**2. Make the brand helper fail loudly instead of leaking quietly.** `withAgentBrandValues` captures `BinaryName` and `ProjectDirName`; the brand package has twelve such variables. A `mutate()` touching any of the other ten restores nothing, and the change follows into every later test in the package, with no failure where it was introduced. Rather than maintain the list by hand — the same bookkeeping problem one step removed — the restore now compares `brand.RuntimeValues()` before and after and fails when they differ. `Values` is a struct of strings, so the comparison is exact and covers fields added later without anyone remembering to.

Its doc comment also records why the helper is safe: no test in this package runs in parallel, and none outlives itself. Both conditions are load-bearing, and the second was broken until the first commit here.

## Change Map

| File | Change |
|------|--------|
| `internal/agent/acpx_agent_test.go` | cancellable context + `t.Cleanup` that cancels and waits; first-progress timeout 2s → 30s |
| `internal/agent/checkpoint_summary_test.go` | restore verifies the whole brand struct and fails on anything left mutated; doc comment records the invariant |

## Validation

- `go test -race ./internal/agent/...` — whole package, 0 races, 0 failures (211s).
- The restore guard was checked by mutating an uncaptured field on purpose: the test fails with `brand values not restored after the test` and names the difference.

Full disclosure on where that ran: this was developed and measured on Windows, on a branch that also carries native-Windows test-harness fixes, because on `main` the fake `acpx` is installed with `os.WriteFile` and Windows will not execute an extensionless script — the test cannot reach its subject there at all. On this branch I verified `go vet` and that the package's tests compile; the behavioural runs are from the ported branch. Both changes are platform-independent and should be quick to confirm on Linux.

## Risks

Low, test-only. If `Run` ever stopped honouring context cancellation, the cleanup would block — surfacing as a timeout naming this test, rather than as a race elsewhere.

## Notes

Touches `internal/agent/acpx_agent_test.go`, as does #103, but a different test — no overlap expected beyond whichever lands second rebasing.

### DIFF
--- a/internal/agent/acpx_agent_test.go
+++ b/internal/agent/acpx_agent_test.go
@@ -544,12 +544,12 @@ func TestACPXAgentRunStreamsOutputToLogsEventsAndProgress(t *testing.T) {
 
 	outputsDir := t.TempDir()
 	progressCh := make(chan struct{}, 4)
-	ctx := withExecutionProgressCallback(context.Background(), func() {
+	ctx, cancelRun := context.WithCancel(withExecutionProgressCallback(context.Background(), func() {
 		select {
 		case progressCh <- struct{}{}:
 		default:
 		}
-	})
+	}))
 
 	var mu sync.Mutex
 	var events []LLMAgentEvent
@@ -564,7 +564,10 @@ func TestACPXAgentRunStreamsOutputToLogsEventsAndProgress(t *testing.T) {
 		err    error
 	}
 	done := make(chan runResult, 1)
+	var runDone sync.WaitGroup
+	runDone.Add(1)
 	go func() {
+		defer runDone.Done()
 		result, err := NewACPXAgent(outputsDir).Run(ctx, LLMAgentRunRequest{
 			BackendName: "codex-acp",
 			AgentID:     "coder-1",
@@ -576,11 +579,23 @@ func TestACPXAgentRunStreamsOutputToLogsEventsAndProgress(t *testing.T) {
 		done <- runResult{result: result, err: err}
 	}()
 
+	// Every assertion below ends the test on the spot when it fails. Without
+	// this the agent goroutine outlives the test, and it reads process-wide
+	// brand values that a later test in this package mutates and restores —
+	// a data race reported against that unrelated test, far from its cause.
+	t.Cleanup(func() {
+		cancelRun()
+		runDone.Wait()
+	})
+
 	select {
 	case <-progressCh:
 	case result := <-done:
 		t.Fatalf("Run() completed before streaming progress; result=%+v err=%v", result.result, result.err)
-	case <-time.After(2 * time.Second):
+	case <-time.After(30 * time.Second):
+		// Generous because this waits on a real process to start and stream:
+		// the previous two seconds were enough when the machine was idle and
+		// not when the rest of the suite was running beside it.
 		t.Fatal("timed out waiting for streamed ACPX progress")
 	}
 
--- a/internal/agent/checkpoint_summary_test.go
+++ b/internal/agent/checkpoint_summary_test.go
@@ -17,9 +17,10 @@ import (
 //
 // Those variables are process-wide: the change is visible to every goroutine,
 // not just this test. That is safe here because no test in this package runs in
-// parallel and none outlives itself — a goroutine still reading brand values
-// after its test returned would race with the restore below, and the detector
-// would report it against whichever test happened to be running at the time.
+// parallel and none that reads brand values outlives itself — a goroutine
+// still reading brand values after its test returned would race with the
+// restore below, and the detector would report it against whichever test
+// happened to be running at the time.
 //
 // Only BinaryName and ProjectDirName are captured, since those are what the
 // callers change. Rather than track that list by hand — a mutation of any other
@@ -34,6 +35,9 @@ func withAgentBrandValues(t *testing.T, mutate func()) {
 	t.Cleanup(func() {
 		brand.BinaryName = oldBinaryName
 		brand.ProjectDirName = oldProjectDirName
+		// A field left as "" rather than mutated is masked here: RuntimeValues()
+		// runs withDerivedDefaults(), which refills empty fields from
+		// NameLower, so a blanked field compares equal and this check misses it.
 		if after := brand.RuntimeValues(); after != before {
 			t.Errorf("brand values not restored after the test:\n got  %+v\n want %+v\n"+
 				"mutate() changed a field this helper does not capture — add it, or the change leaks into the tests that follow",
--- a/internal/agent/checkpoint_summary_test.go
+++ b/internal/agent/checkpoint_summary_test.go
@@ -12,14 +12,33 @@ import (
 	"github.com/liza-mas/liza/internal/testhelpers"
 )
 
+// withAgentBrandValues applies mutate to the brand variables and restores them
+// when the test ends.
+//
+// Those variables are process-wide: the change is visible to every goroutine,
+// not just this test. That is safe here because no test in this package runs in
+// parallel and none outlives itself — a goroutine still reading brand values
+// after its test returned would race with the restore below, and the detector
+// would report it against whichever test happened to be running at the time.
+//
+// Only BinaryName and ProjectDirName are captured, since those are what the
+// callers change. Rather than track that list by hand — a mutation of any other
+// field would leak into every later test in the package, silently — the restore
+// checks the whole struct and fails if anything is left behind.
 func withAgentBrandValues(t *testing.T, mutate func()) {
 	t.Helper()
+	before := brand.RuntimeValues()
 	oldBinaryName := brand.BinaryName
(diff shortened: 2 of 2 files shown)


tangi-vass (dismissed):
Reviewed at `6be8df187e9a295a7ac4d0d58476da3d045fa691`.

**Review: Standard — Approve.** Blockers: none. Concerns: none. Suggestions: 2.

The diagnosis is right and the fix matches it. Three things I checked rather than took on trust, because each is what makes the cleanup safe rather than a new hang:

- `done` is buffered (`make(chan runResult, 1)`), so the abandoned goroutine's send cannot block and `runDone.Wait()` cannot deadlock on it.
- The `t.Cleanup` is registered before the first statement that can call `t.Fatal`, and cleanups run LIFO — so it runs *ahead* of the `t.Setenv("PATH", …)` restore and the `t.TempDir` removals registered earlier. The goroutine still has its PATH and its directories while it unwinds.
- Every process launch on this path uses `exec.CommandContext` (`acpx_agent.go:175`, `:260`, `:273`), so `cancelRun()` actually reaches the child.

## Validation

Ran on Linux in a disposable worktree at this head, since CI has not:

- `go test -race -count=1 ./internal/agent/...` — 844 passed, 4 failed, 1 skipped, **0 data races**. The 4 failures are all in `launch_resolver_test.go` and reproduce identically at the merge base `c0eff8fa`, in isolation — pre-existing on this machine, unrelated to this PR.
- `go vet ./internal/agent/...` — clean.
- Reproduced the guard independently: adding `brand.NameUpper = "ACME"` to an uncaptured `mutate()` fails with `brand values not restored after the test` and prints the differing field. It does what the commit message says.

Note for whoever merges: CI run `31247331264` concluded `action_required` — fork PRs need a maintainer to approve the workflow, so no check has run against this head. `protect-main` requires one approving review and no status checks, so nothing is formally blocked, but the workflow is worth approving before merge.

## Findings

**[suggestion]** `internal/agent/checkpoint_summary_test.go:37` — the guard has one blind spot: a field left as `""`.

`RuntimeValues()` runs `withDerivedDefaults()`, which refills empty fields from `NameLower`. So a `mutate()` that *blanks* an uncaptured field compares equal and passes. Verified at this head: with `brand.NameUpper = ""` inside a caller's `mutate()`, the test reports `ok` while `brand.NameUpper` stays empty for every test that follows — exactly the silent leak this commit exists to stop.

Likelihood is low but not contrived: the test most likely to blank a brand field is one exercising the derivation defaults themselves, which is normal work in a brand-aware package.

Cheapest close is the comment rather than the code — comparing the raw package vars means listing the twelve by hand, which is the bookkeeping you deliberately avoided. Saying "fields left empty are masked by the derived defaults" keeps the next reader from over-trusting it.

Closure condition: none — non-blocking, your call whether it is worth the line.

**[suggestion]** `internal/agent/checkpoint_summary_test.go:19-20` — "none outlives itself" reads as a package-wide invariant, and as written it is not one.

`heartbeat_test.go:236` and `waitforwork_test.go:610` both start goroutines that their `t.Fatal` paths cancel without waiting for, so goroutines in this package do outlive their tests. I checked whether that matters: neither reaches `brand` at runtime — `heartbeat.go` and the strategies never touch it, and the runtime readers (`paths.ProjectDirName`, `models.CurrentDestructiveDBAllowMarker`) are not on those paths. So the safety claim holds today under the narrower reading, "none *that reads brand values* outlives itself".

That narrower sentence is worth writing down, because the comment is explicitly load-bearing — you flagged it as such — and the broad version is the one a future reader will check their new goroutine against.

Closure condition: none — non-blocking.

**[appraisal]** The PR body names the victim/cause distinction and the reason the report is hard to read, and the second commit generalizes the guard instead of extending a hand-maintained list. The 30s value is also correctly described as a ceiling rather than a sleep — the fake `acpx` returns in ~0.1s, so it costs nothing on the passing path.

## Net value

As submitted: removes a real cross-test data race and the goroutine leak that causes it, for 37 test-only lines — clearly better than not merging.
With findings resolved: the same, plus a guard whose documented coverage matches what it actually catches.

Blast radius: Low — test-only, no production code touched.
Confidence: high — both files read in full at this head, race suite and guard behaviour verified locally.
Sources: PR body and both commit messages; full diff; `acpx_agent_test.go` and `checkpoint_summary_test.go` at head; `internal/brand/brand.go`; `acpx_agent.go`; `heartbeat.go`, `waitforwork.go` and the strategies for the brand-reachability check; `.github/workflows/ci.yml`; the `protect-main` ruleset; local `-race` and `vet` output at this head and at the merge base.

Next step: merge — approve the CI workflow run first so the change gets one clean pass on ubuntu and macos.

tangi-vass:
Reviewed at `989e7cad09faa025b82441c22cd5292f12e845a2`.

## Review: Standard — Comment

Blockers: None in code; approval is blocked by one required check  
Concerns: None  
Suggestions: None

Overall: The goroutine cancellation/join fixes the reported cross-test race without changing production behavior. The brand-helper documentation now accurately states both known limitations.

Approach: Sound. Merely increasing the timeout would reduce the failure frequency but would not prevent assertion paths from abandoning the goroutine; cancellation plus an explicit join addresses the root cause.

Absence: Both changes declared in the PR body are present. No code, test, or documentation deliverable is missing from that scope.

## Prior review reconciliation

- **RESOLVED:** The package-wide “none outlives itself” claim now applies only to goroutines reading brand values.
- **RESOLVED:** The comment now documents that `RuntimeValues()` masks fields left empty through derived defaults.

## Validation

- Exact-head Ubuntu and macOS CI succeeded. The workflow ran `make lint`, race-enabled `make test`, `make test-e2e`, and `make build`.
- In a disposable exact-head worktree, I forced the streaming test’s pre-progress timeout path. After the documented `make sync-embedded` prerequisite, the test reached its intended assertion and exited normally rather than hitting Go’s five-second harness timeout. This confirms cleanup cancels and joins the running agent goroutine.

## Required check

The active `protect-main` ruleset requires `license/cla`. Its exact-head status is currently `pending`: “Contributor License Agreement is not signed yet.”

Closure condition: `license/cla` succeeds for this head, followed by another review-state refresh.

Net value as submitted: removes a real goroutine leak and cross-test data race for a small, test-only change.  
Net value with the external blocker resolved: positive and ready for approval.

Blast Radius: Low — test-only.  
Confidence: High — complete diff, relevant exact-head sources, prior review history, CI configuration/results, branch ruleset, and failure-path behavior inspected.  
Next step: sign or recheck the CLA, then request approval again.

PR #107, opened 2026-08-10
feat(windows)!: restore native Windows support

jgicquel:
Reopens the question closed in v0.8.0 (`788a9f4`, "drop native Windows support, use WSL2"), and reports what porting actually cost. Discussed in #104.

## Why

WSL2 is unavailable on locked-down corporate machines — no virtualisation, no admin rights to enable it. That is the environment this branch was built and measured in, and it is a common one.

The question this PR really answers is not "can Liza run on Windows" but **"what did Windows expose that was already wrong?"** — because most of what follows are platform-independent defects that only Windows made visible.

## Defects found in production code

None of these were introduced by this branch. Each was reached by a real failure, not by inspection.

| Defect | Effect before the fix |
|---|---|
| `resolveWorktreePath` joined the project root and the stored relative path with a literal `/` | The mixed path is the target root handed to Semble, which compares it against the expected worktree root and refuses — **the entire Semble section vanished from Windows prompts, silently** |
| `MergeWorktree` exec'd `scripts/integration-test.sh` directly | Windows cannot fork/exec a POSIX script: **a project's integration tests never ran at merge**, and the pipeline decided on an empty `TestOutput` |
| `launchShell()` returned `/bin/sh` when `SHELL` is unset | `SHELL` is always unset under Git for Windows, so **`liza launch wezterm` was inoperable** — hidden because the tests set `SHELL` to a stub |
| `session-context.sh` classified an absolute path as relative unless it began with `/` | git returns `core.hooksPath` verbatim when it is absolute, so on Windows the hook prefixed the project dir to `C:\...`, found nothing, and **skipped every index section** |
| `session-context.sh` canonicalised `cwd` but not `CLAUDE_PROJECT_DIR` | The same session emitted `C:\Users\proj/stacklit.json` or `C:/Users/proj/stacklit.json` depending only on which source won |
| `reconcilePreviousProviderActivations` used `os.UserHomeDir` where the rest of the flow uses `paths.UserHomeDir` | On Windows the first reads `USERPROFILE` and ignores `HOME`, so reconciliation compared against a different home than the one the activation was placed in — and **in tests, edited the developer's real `~/.codex`** |
| `plugin/acp` measured run duration across `time.Now().UTC()` | Windows advances its clock in ~15ms steps, wall and monotonic alike, so a fast run measured as exactly zero |
| `make install` resolved `~/.local/bin` from `HOME` | `HOME` is not a Windows variable; from PowerShell the install directory became `/.local/bin`. With that fixed, `mkdir -p` still failed — Make runs a metacharacter-free recipe line through CreateProcess without a shell, and the POSIX utilities live in `Git\usr\bin`, deliberately off PATH |

Two further defects were platform-independent and are proposed separately in #106 (a leaked agent goroutine racing with process-wide brand values).

## Approach

Three decisions shaped everything else:

**Git for Windows is required.** Hooks stay POSIX and run through `bash`. Porting them to PowerShell would double the surface the behavioural contract has to hold, for no gain.

**Real symlinks, as on Unix.** Developer Mode or an elevated shell is needed to create them. Tests that need one go through `testhelpers.RequireSymlinkCapability`, which *probes the capability* rather than branching on `GOOS` — so they run for real wherever the privilege is held, including Windows CI, and skip with an actionable message only where they genuinely cannot run.

**No convenience skips.** A skip is acceptable only where the concept does not exist on Windows. Where a POSIX mechanism has a Windows equivalent, the equivalent is implemented — see `internal/perm` (chmod ⟷ explicit DACL deny) and `testhelpers.PreventRemoval` (chmod ⟷ open handle, because Windows will not let a process deny itself deletion of a path it owns).

## Change map

- **Paths** — one canonical form for values that travel between machines (`state.yaml`, catalog YAML: slash), native form for values a tool consumes. Path validation, path escaping in errors, and launch-path comparison follow.
- **Hooks** — Git Bash resolution, absolute-path recognition, separator canonicalisation, `.cmd` wrapper install as a first-class path, Go pre-commit hooks made runnable.
- **Process** — spawned agents detached from the parent console group, so Ctrl+C on the TUI no longer reaches them.
- **Toolchain** — winget/scoop/choco with verified qualified IDs, `.exe` renaming with post-install verification, `env.ps1` alongside `env.sh`.
- **Test harness** — stubs installed as a real relay executable (a `.cmd` wrapper forwarding `%*` cannot carry a newline inside one argument, so any multi-line payload arrived truncated at its first line); git fixtures pinned to `core.autocrlf=false`; two seams where Windows cannot produce a non-ENOENT stat failure.
- **Build/CI/docs** — `make install` fixed, build version derived from git, `-timeout 30m` on the test target, Windows sections in the troubleshooting guide, ADR.

## Validation

`make test` (which is `go test -race`) on windows/amd64, with Developer Mode enabled, on an otherwise idle machine:

```
49 packages ok · 0 failing · 0 data races · 0 timeouts · ~25 min
```

Two commits landed after that full run — a `make install` fix and a test-cleanup fix — and were each verified with three consecutive `-race` runs of the affected tests rather than another full sweep.

Honest caveats, in order of importance:

1. **Developer Mode must be on**, otherwise every symlink-gated test skips and the suite is green without having exercised contract linking. The measurement above was taken with it enabled.
2. **`-timeout 30m` is load-bearing on Windows.** `internal/ops` takes ~15 min and `internal/commands` ~10 min under `-race`; the 10m default kills the package mid-run, and — this is the trap — the tests after the one running at that moment are never reported, so a truncated measurement reads like a handful of failures rather than an incomplete run.
3. A saturated `%TEMP%` (a few hundred leftover test directories) doubles `internal/ops` runtime and pushes timing-sensitive tests over their budgets. Worth clearing before judging a Windows run. More generally, a handful of tests here compare durations or wait on a deadline, and a busy machine tips them over: on a loaded box I saw two to four of them fail per sweep, each passing on three consecutive isolated `-race` runs. Worth knowing before reading a red Windows sweep as a code defect — it is the trap this branch fell into repeatedly.
4. Linux and macOS are unchanged by construction — every path change is either `filepath.Join` where a literal `/` was used, or an expectation aligned with what the code already produced off Windows — but I have no way to run those here. **That is the review I would most value.**

## Not in scope

- `/tmp` in `internal/embedded/claude-settings.json` has no Windows equivalent. Left alone: a machine-specific path in a committed file would be worse. Raised for discussion.
- `scip-python` is broken on Windows independently of Liza (`new RegExp(path.sep, 'g')` yields `/\/g`), so Python SCIP indexing cannot work until that is fixed upstream of us.
- Two tests reach the network to fetch `provider-catalog.yaml` from `main` and compare it against hardcoded expectations. Any branch not resynchronised the moment that file changes goes red for reasons unrelated to its own content — it happened to this branch mid-review.

## Reviewing this

56 commits, each one scoped to a single defect with the measurement that exposed it in the message. Reading them in order is probably easier than reading the diff: the fixes are individually small, and the interesting part is usually *why* the failure was invisible until Windows.

Happy to split this into a sequence of smaller PRs if you prefer — the hook fixes stand alone, and the test-harness work is separable from the production fixes.


tangi-vass:
Reviewed at `cd2429d273f40a8446271aaacadc7fa0a0d859fc`. Standard review of the full change set, deepened on the production paths (paths, worktrees, process, hooks, toolchain, updater, install).

**First, the review you asked for.** I ran `make test` (`go test -race`) on linux/amd64 at this head, in a clean worktree pinned to the exact SHA:

```
47 packages ok · 2 packages FAIL · 0 data races
```

All three failing tests are accounted for, and none is attributable to this branch:

| Test | Verdict |
|---|---|
| `TestNonDefaultBrandBuildSmoke` | My harness. `error obtaining VCS status: exit status 128` — Go's VCS stamping does not work in the linked worktree I reviewed from. A plain `go build ./cmd/liza` fails the same way there, and `-buildvcs=false` builds clean. This branch does not touch buildvcs. |
| `TestExecuteAgentOutputProgressPreventsProgressTimeout` | Load-sensitive. Passes 3/3 isolated `-race` runs. Not touched by this branch. |
| `TestLizaSimplifiedProjectACPPerformanceComparison` | Load-sensitive (`796ms` vs `769ms` baseline). Passes 3/3 isolated `-race` runs. Not touched by this branch. |

So: **no Linux regression, and no data race.** Your caveat 3 reproduced exactly as written — two timing-sensitive tests failed under a full parallel `-race` sweep and passed on three consecutive isolated runs each. That prediction being precise is itself worth something; it is the difference between a caveat and an alibi.

Note that CI has never run on this head: the workflow run is `action_required` (first-time-contributor gate), so nothing here has been validated by the repo's own pipeline. The `protect-main` ruleset requires one approving review and no status checks, so this does not formally block — but it means the Windows job this PR adds is still unexercised.

---

## [blocker] `install.sh:208` — the Git Bash install path aborts after downloading

`detect_platform` now returns `os="windows"` for `MINGW*|MSYS*|CYGWIN*` instead of refusing, so `install_release` runs on Windows and correctly fetches the `.zip`. But `BINARY_NAME` is `liza` throughout the script, and the Windows archive contains `liza.exe` — goreleaser builds `binary: liza` and Go appends `.exe` for `GOOS=windows`. So:

```sh
chmod +x "${tmp_dir}/${BINARY_NAME}"   # ${tmp_dir}/liza does not exist
```

`chmod` exits 1, `set -e` is in force (line 5), and `install_release` is invoked as a plain command at line 334 — nothing masks it. The script dies there, after the download and checksum verification, with a bare `chmod: cannot access` and no installed binary.

The two lines after it carry the same assumption: the verify step at 212 and the `mv` at 227/230 would install a file named `liza` with no extension, which Windows will not resolve through `PATHEXT` — the exact failure `verifyInstalled` in `internal/toolchain/install.go` was written to repair for tools.

**Likelihood:** every run. This is the path a Git Bash user reaches by following the README's `curl | bash`, and it is newly reachable *because* of this change — before it, the script refused with an actionable message.

**Why the tests miss it:** `runInstallScriptHelp` only ever invokes `install.sh --help`. Nothing exercises download → extract → install.

The inline comment on 207 — `# Make executable (no-op on Windows; harmless under Git Bash)` — suggests the line was considered and judged on `chmod`'s semantics rather than on whether the filename resolves. That is the only reason I am confident this is an oversight rather than a decision.

**Closure condition:** either `install.sh` installs a runnable `liza.exe` under Git Bash, or it declines Windows explicitly and points at `install.ps1`. Both are small; I have no preference between them, and the second may well be the honest one given `install.ps1` exists and is the documented Windows path.

---

## [concern] `cmd/liza/cmd_launch.go:1070` — `launchShell()` resolves the WSL launcher on the machines this PR targets

```go
if runtime.GOOS == "windows" {
    if bash, err := exec.LookPath("bash"); err == nil {
        return bash
    }
}
```

`exec.LookPath("bash")` on a stock Windows box finds `C:\Windows\System32\bash.exe` — the WSL launcher — unless Git for Windows is ahead of `system32` on the **machine** PATH. You document precisely this in `support-docs/TROUBLESHOOTING.md:462`, including that a user-level PATH entry *cannot* win and the fix needs an elevated machine-PATH change.

That is the bind: the PR's premise is corporate machines with no admin rights and no WSL. On such a machine `System32\bash.exe` is present, `LookPath` returns it, and `liza launch` hands a POSIX pane script with `C:/...` paths to a launcher that cannot see them — and cannot be fixed without the elevation the user does not have.

You already solved this: `testhelpers.ResolveBashForScripts` probes `LOCALAPPDATA\Programs\Git\bin`, `ProgramFiles\Git\bin`, `ProgramFiles(x86)\Git\bin` before falling back to `LookPath`. The production path skips that discipline. The `check-testhelpers` gate stops you importing the helper directly, so this needs the candidate list to live production-side — but it is the same twenty lines.

`internal/ops/shell_runner_windows.go`'s `LookPath("sh")` is not exposed to this particular trap (`system32` ships no `sh.exe`), so the fix is local to `launchShell`.

**Closure condition:** `launchShell()` on Windows prefers a Git for Windows `bash.exe` over a `system32` match, or documents why the launcher is acceptable here.

---

## [concern] `internal/procscan/process_alive_windows.go:19` — `OpenProcess` alone reports exited processes as alive

A handle can still be opened for a terminated process as long as any process holds a handle to it — `OpenProcess` succeeds and `ProcessAlive` returns `true`. The canonical probe pairs it with `GetExitCodeProcess` and treats anything other than `STILL_ACTIVE` as dead.

This feeds `IsProcessAlive` in `internal/ops/delete_agent.go` and `signalProcessStatus`, so the failure mode is a dead agent held as a live row — which the doc comment on the `default` branch names as the thing to avoid ("be conservative and report dead rather than holding a phantom agent row"). The `default` branch is conservative; the success branch is not.

I have not measured this against a real Liza spawn, so I cannot tell you how often a handle survives in practice — it depends on whether the spawning process outlives the agent without reaping it. That uncertainty is why this is a concern and not a blocker.

**Closure condition:** either `GetExitCodeProcess` gates the alive result, or a note records why a surviving handle cannot occur for spawned agents.

---

## [suggestion] `internal/toolchain/install.go:405` — the first present package manager ends the search

```go
packageName, ok := packageNameFor(tool, candidate.binary)
if !ok {
    return Command{}, errNoPackagePath
}
```

`return`, not `continue`. On a host with winget (i.e. any modern Windows) the loop stops at winget for any tool lacking a winget entry, without consulting scoop or choco — even when they are installed and the catalog names the tool for them.

Today's catalog makes this unreachable: every `InstallPackage` tool either has all three Windows IDs (`rg`, `ast-grep`, `jq`, `yq`, `gh`) or none (`mdq`, `pre-commit`, both with a `ManualInstallNote`). So this is latent, not live. It becomes live the first time someone adds a scoop ID without a winget one — which is the normal way a catalog grows, since scoop carries things winget does not.

`continue` preserves the stated intent ("stop rather than guess") while letting a manager that *does* know the tool answer.

## [nit] `internal/procscan/process_alive_windows.go:33` — not gofmt-clean

The `const` block's alignment is off by two columns. `make lint` runs `go fmt`, which rewrites rather than fails, so CI will not catch it — which also means the pre-commit `go-fmt` hook did not run over this file.

## [question] `internal/toolchain/install.go:265` — `LIZA_*` env vars vs the branding guardrail

`windowsArchiveCommand` introduces `LIZA_TOOL_ARCHIVE_URL`, `LIZA_TOOL_BINARY`, `LIZA_TOOL_INSTALL_DIR` as literals. `GUARDRAILS.md` G1.3 routes user-visible brand identifiers through `internal/brand`, with an allowance for legacy `LIZA_*` aliases.

I am flagging rather than asking for a change: the neighbouring `apt-get` branch has used a literal `LIZA_TOOL_PACKAGE` since before this PR, and matching the line above you is the right call absent a maintainer ruling. Worth a maintainer answer on whether the guardrail intends to cover this class at all — these never surface to a user, they only cross into a spawned shell.

## [appraisal]

Two things worth naming.

`RequireSymlinkCapability` probes the capability by creating a symlink rather than branching on `GOOS`. That inverts the usual shape — the tests run for real wherever the privilege exists and skip only where they genuinely cannot, instead of being switched off by platform. Combined with your caveat 1 (Developer Mode off ⇒ a green suite that never exercised contract linking), it is the difference between a port and a port you can trust.

And the `.cmd`-wrapper diagnosis in `stubrelay`: "cmd.exe cannot carry a newline inside a single argument, so a multi-line payload arrives truncated at its first line." That is the kind of finding that costs a day and reads like one line afterwards.

---

## Review summary

**Review:** Standard (deepened on production paths) — **Comment**

- **Blockers:** 1
- **Concerns:** 2
- **Suggestions:** 1 (+1 nit, 1 question)

**Overall:** The framing is right — most of this is platform-independent defects that Windows made visible, and the two production fixes with the widest blast radius (`resolveWorktreePath` handing Semble a mixed-separator root; `MergeWorktree` exec'ing a POSIX script it could not run) are real bugs on every platform. The blocker is a one-line consequence of newly enabling a path that was previously refused.

**Approach:** Sound. The three stated decisions (Git for Windows required, real symlinks, no convenience skips) are the ones that keep the behavioural contract single-surfaced; porting hooks to PowerShell would have doubled it. The alternative I weighed was `GOOS`-gated skips for the symlink and permission tests — cheaper, and it would have produced a green Windows suite that proved nothing. Capability probing is the more expensive and correct rung.

**Absence:** Checked against your Change map and Not-in-scope declarations. Docs, ADR (0112, registered in the ADR index), CI job, and troubleshooting entries are all present. Stale WSL2 guidance is gone from `install.sh`, `internal/toolchain/install.go` and `GETTING_STARTED.md`; the surviving WSL references are all deliberate warnings about the launcher. What is missing is any test over `install.sh`'s install path — which is what let the blocker through.

**Net value:** *As submitted* — restores a platform and fixes eight real defects, but ships a documented install path that cannot complete and a `liza launch` that misfires on precisely the locked-down machines motivating the work. *With the findings resolved* — clearly positive: the production fixes stand on their own merit off Windows, and Linux is measured unchanged.

**Blast radius:** High — build, install, distribution, process spawning, path canonicalisation, git hooks.

**Confidence:** Medium-high. The Linux suite, the blocker, and the gofmt finding are measured. The two Windows concerns are read from source and platform semantics — I have no Windows host, so `[concern]` reflects that, not a hedge about their reasoning.

**Sources:** full PR diff (107 files) via the files API; source at `cd2429d2` in a detached worktree for `install.sh`, `.goreleaser.yaml`, `Makefile`, `internal/toolchain/*`, `internal/procscan/*`, `internal/ops/wt_merge.go`, `internal/perm/*`, `internal/testhelpers/utils.go`, `support-docs/TROUBLESHOOTING.md`; `make test` output; targeted isolated `-race` reruns; `gofmt -l`; repo ruleset and check-run state; `GUARDRAILS.md`.

**On splitting:** you offered to. I would not, for the production fixes — they are individually small and the commit-per-defect history is genuinely readable, which is the thing a split usually buys. If anything is separable it is the test-harness work (`stubrelay`, `perm`, `removal`, `filemode`), which stands alone and is the largest single block of new surface. Your call; it is not a review finding.

tangi-vass:
Review: Deep — Request Changes

Blockers: 4 total — 2 new below, 2 confirmed from the earlier review
Concerns: 3 total — 1 new below, 2 existing
Suggestions: 1 existing; none new

Overall: The Git Bash compatibility strategy and capability probing are appropriate, but required native-Windows initialization and agent-lifecycle paths remain broken. Existing review findings also leave release installation and launch-shell selection unresolved.

Reconciliation: The [earlier independent review](https://github.com/liza-mas/liza/pull/107#issuecomment-5242390046) already reports the `install.sh` failure, `launchShell` selection risk, `ProcessAlive` concern, and formatting defect. I independently confirmed the installer failure and formatting rewrite, so I am not duplicating those findings. I classify the formatting defect as blocking because the exact-range repository pre-commit command deterministically fails and modifies the checked-in source.

Approach: Sound at the system level. Keeping one hook implementation under Git Bash avoids duplicating initialization policy in PowerShell; the remaining problems are missing adaptations at Windows path, lifecycle, and distribution boundaries.

Absence: The PR description, issue #104, and ADR-0112 are the behavioral authority for native Windows support. Regression coverage is missing for `CLAUDE_PROJECT_DIR`-backed native reads and production Windows agent termination.

Net value: As submitted — negative versus not merging, because several advertised native-Windows paths can fail or silently orphan processes. / With all findings resolved — positive, restoring native Windows support while retaining the existing Linux/macOS implementation and useful cross-platform fixes.

Blast Radius: High — public platform support, installation, initialization enforcement, and agent lifecycle behavior.
Confidence: High for the reported findings: the complete 107-file diff and exact-head sources were reviewed and locally validated; no Windows runtime host was available, but both new blockers have deterministic source paths.
Sources: PR description; issue #104; ADR-0112; complete base-to-head diff; exact-head source and tests; project guardrails and architecture records; current comments, reviews, threads, branch rules, and check status; exact-head local validation.
Next step: Resolve all remaining blockers and concerns, add the missing boundary regressions, and request another review.

Round 1 — new blockers 0→2, files 107→107

[blocker] internal/embedded/hooks/enforce-init.sh:117 — `CLAUDE_PROJECT_DIR` remains in the native Windows path namespace

Why it matters: The hook normalizes Read paths from `C:\...` to `C:/...`, but assigns `project_dir` afterward and does not normalize it. Absolute required-document reads therefore do not match patterns such as `"$project_dir"/GUARDRAILS.md`, so the initialization gate can remain active after every required document was read.

Likelihood: The runtime supplies `CLAUDE_PROJECT_DIR` as a native path on Windows. The sibling `session-context.sh` explicitly handles this case, while `TestEnforceInitHook_NativeReadsClearPairingGate` does not set that environment variable and therefore misses it.

Suggestion: Normalize `project_dir` immediately after resolving it, using the same Windows-only conversion as `session-context.sh`, and add a test combining a native `CLAUDE_PROJECT_DIR` with absolute native Read paths.

Closure condition: On Windows, absolute native reads of all required Pairing documents clear the gate when `CLAUDE_PROJECT_DIR` is set to a native path.

[blocker] internal/ops/delete_agent.go:138 — Running Windows agents are silently left alive while their state is deleted

Why it matters: `terminateProcess` returns a successful no-op when `isLizaAgent` is false. The production identity check reads `/proc/<pid>/cmdline` and explicitly returns false on platforms without procfs, so Windows never signals the process. `TerminateAgent` then deletes the state row and the CLI reports `Deleted agent`, leaving the agent or provider process orphaned.

Likelihood: Any eligible deletion of a running Windows agent, whether forced or interactively confirmed, reaches this production boundary. Unit tests replace `agentProcesses`, so they do not exercise the Windows behavior.

Suggestion: Add a Windows process-identity implementation and termination behavior that covers the spawned process tree. If identity cannot be verified safely, fail closed and retain the state row.

Closure condition: A Windows behavioral test demonstrates that deleting a registered running agent leaves neither its recorded process nor its spawned provider alive before removing state, and that an unverifiable or reused PID is not signaled or silently discarded.

[concern] install.ps1:3 — The new distribution surface bypasses the white-label boundary

Why it matters: The PowerShell help text exposes raw default-brand literals in the synopsis, install path, executable name, and examples even when `BRAND_*` values select another brand. This violates project guardrail G1.3 for end-user-visible distribution surfaces.

Likelihood: `Get-Help` or direct inspection of `install.ps1` in every non-default-branded distribution shows the default brand. The centralized fallback values and canonical upstream URL are allowlisted; the user-facing help text is not.

Suggestion: Make the help brand-neutral or render it during release generation, and include `install.ps1` in non-default-brand distribution validation.

Closure condition: A non-default-brand render/help scan contains no raw default-brand identifiers outside the documented allowlist.

Validation:
- ✅ `git diff --check`
- ❌ Exact-range `pre-commit`: `go-fmt` rewrote `internal/procscan/process_alive_windows.go`
- ⚠️ `make lint`: exited 0 but rewrote the same source file
- ✅ `make test` with race detection
- ✅ `make build`
- ✅ `go build ./...` for Windows amd64/arm64 and Darwin amd64/arm64
- ⚠️ GitHub exposes no CI checks for this exact head

jgicquel:
Corrective round 1 — `cd2429d2` → `84c5f561`, 8 commits, +890/−90 across 20 files.

One commit per finding. Thanks to both reviewers: the two blockers I could not
have found from Linux were the two that mattered most, and the `ProcessAlive`
concern turned out to be measurable rather than hypothetical.

## Findings

| # | Finding | Status | Commit |
|---|---|---|---|
| 1 | [blocker] `install.sh:208` — install path aborts after downloading | RESOLVED | `20e80dd6` |
| 2 | [blocker] gofmt on `process_alive_windows.go` | RESOLVED | `68bd46e6` |
| 3 | [blocker] `enforce-init.sh:117` — `CLAUDE_PROJECT_DIR` not normalised | RESOLVED | `2788560b` |
| 4 | [blocker] `delete_agent.go:138` — Windows agents left alive | RESOLVED | `84c5f561` |
| 5 | [concern] `cmd_launch.go:1070` — `launchShell` resolves the WSL launcher | RESOLVED | `338d5bab` |
| 6 | [concern] `process_alive_windows.go:19` — `OpenProcess` alone | RESOLVED | `535218f9` |
| 7 | [concern] `install.ps1:3` — brand literals in the help text | RESOLVED | `2c9cd21c` |
| 8 | [suggestion] `install.go:405` — first present manager ends the search | RESOLVED | `bead9f15` |
| 9 | [question] `LIZA_TOOL_*` vs G1.3 | left to the maintainer | — |

### 1 — `install.sh` declines Windows

Option (b), as suggested: `detect_platform` refuses and points at `install.ps1`,
which `GETTING_STARTED.md` already documents as the Windows path. The zip/unzip
branches are gone, so `install_release` is byte-identical to its pre-branch shape.

One thing worth reporting, because it also applies to the code this replaces:
`detect_platform`'s stdout is consumed by `platform=$(detect_platform)`, so a
refusal printed there goes into a variable rather than to the user. The original
`Error: Native Windows is not supported.` had the same blind spot. The new refusal
writes to stderr, and the test asserts on process output, so moving it back to
stdout fails the test.

Verified under Git Bash on Windows (`uname -s` = `MINGW64_NT-10.0-26200`): refuses
with exit 1 before any download, and renders with zero default-brand literals under
a `BRAND_NAME_LOWER` / `BRAND_REPO` override.

### 3 — `CLAUDE_PROJECT_DIR`

Normalised immediately after resolution, the same Windows-only conversion
`session-context.sh` applies. Both derived branches were already safe — git reports
a forward-slashed toplevel, and `cwd` had just been rewritten — so only the
environment-supplied value was exposed, which is exactly why the existing test
missed it. `TestEnforceInitHook_NativeReadsClearPairingGate` now runs both ways and
fails on the environment case without the fix.

### 4 — Windows agent termination

Both halves of the closure condition are covered, and both were verified against
real processes rather than substituted `agentProcesses`.

Identity now reads the command line through `NtQueryInformationProcess` with
`ProcessCommandLineInformation`. That answers from the kernel — no
`ReadProcessMemory`, no PEB layout — and needs no more access than the liveness
probe already takes. `golang.org/x/sys/windows` was already a direct dependency, so
nothing new was pulled in.

Termination walks a Toolhelp snapshot and kills deepest first. Snapshot parent PIDs
are not trusted on their own: the field survives the parent's death and Windows
recycles PIDs, so a candidate whose creation time predates its parent is dropped.

There is no graceful stage, and the code says so rather than implying one: no
SIGTERM, and console control events cannot be raised for another process's group
from an unrelated command, so `signalTree` and `killTree` are the same operation.

For the second half — an unverifiable or reused PID — such a process is still not
signalled, but it is no longer silent: `ProcessTerminationResult.IdentityUnverified`
carries it and the CLI warns that something is still running under the PID whose row
it just removed.

Both tests were checked against the defect they cover: restoring the procfs identity
makes the first fail with the agent unsignalled, and terminating only the root
leaves the grandchild alive.

### 6 — `ProcessAlive`

Measurable, as it turns out. `os/exec` holds a handle from `Start` to `Wait` so the
PID cannot be recycled, and every agent is spawned that way — so a child that exits
immediately was reported alive 30 seconds later, and would have been indefinitely.
With `GetExitCodeProcess` gating the result it reads dead within 0.09s.

Two costs are stated in the source. A process exiting with code 259 reads as alive,
which is inherent to the API — asking for `SYNCHRONIZE` so `WaitForSingleObject`
could answer instead would widen the access rights and with them the
`ERROR_ACCESS_DENIED` cases. And a `GetExitCodeProcess` failure reports alive rather
than dead, the opposite trade from the unknown-`OpenProcess`-error branch, because
there the PID is entirely unknown while here only its exit code is.

Worth noting for symmetry: Unix keeps its own semantics, where `signal(0)` succeeds
on a zombie. The two probes are not identical, and the Windows one is now stricter.

## Declared scope extensions

Beyond the files named in the findings, so they are not read as unannounced growth:

- `internal/procscan/liza_agents.go` — `IsLizaAgentArgv` now trims the executable
  suffix. Without it the Windows identity check cannot match, since the image is
  `liza.exe`. Applied unconditionally rather than per-GOOS; a bare `liza` is what a
  Unix `argv[0]` already is.
- `internal/ops/delete_agent.go`, `internal/commands/delete_agent.go` — the
  `IdentityUnverified` field and the CLI warning, required by the "not silently
  discarded" half of finding 4's closure condition.
- `internal/ops/delete_agent_process_unix.go` — `isLizaAgentProcess` moved here from
  the shared file so the Windows build can supply its own. No behaviour change.
- `internal/procscan/process_alive_windows.go` — `processProbeSource` now reports
  `OpenProcess+GetExitCodeProcess`; the old label became untrue.
- `install.ps1` — one comment naming a command by brand, same file and same rule as
  finding 7.
- Tests for the files above, plus `cmd/liza/install_script_test.go`, where
  `runInstallScriptHelp` was generalised so the refusal path can be exercised.

## Reported, not fixed

- **Five `TestHook_*` tests in `internal/ops` fail on Windows, and already failed at
  `cd2429d2`.** Confirmed in a detached worktree at the reviewed head, so this is not
  a regression from these fixes — but it does contradict the all-green Windows claim
  in the PR description, and I would rather say so than let it be discovered. Cause,
  reproduced outside the repository: `commitInIsolatedPath` replaces the whole `PATH`
  with the stub directory from `makeIsolatedPath`, and git then cannot start a
  `#!/bin/sh` hook — the `cannot spawn <hook>: No such file or directory` is ENOENT on
  the interpreter, not on the hook, which is present and executable. Adding the
  directory of `git.exe` is not enough; the shell's directory is what is missing. The
  fix looks like one line in `makeIsolatedPath`, but it is outside every finding here,
  so it is not in this round.
- `DeleteAgentResult.SignalProcess()` carries the same identity defect and sends
  `SIGTERM`, but is called only from its own tests — dead code in production.
- `install.sh`'s `*)` unsupported-OS branch prints to stdout, so its message is
  swallowed by the same command substitution described in finding 1. Pre-existing and
  outside the finding, so left alone.

## Validation

On Windows, with `%TEMP%` cleared beforehand: `internal/commands` (464s),
`cmd/liza` (290s), `internal/embedded` (280s), `internal/procscan` and
`internal/toolchain` all green. `internal/ops` green apart from the five
`TestHook_*` above. `go build ./...` for linux and darwin on amd64/arm64 and for
windows/arm64. Exact-range pre-commit clean over all 20 touched files, including the
`go-fmt` hook that finding 2 was about.

Linux and macOS remain untested here, so the request for review on those stands.

jgicquel:
Correction to the last item of my previous comment: the five `TestHook_*` failures
are fixed, in `cfabe1ca`, and I got their provenance wrong.

**What they were.** The hook-chain tests replace `PATH` with a directory holding only
their stubs, so the hook sees a controlled set of binaries. On Windows that also
takes away the shell git needs to start a `#!/bin/sh` hook, so the commit failed
before the hook ran at all. The `cannot spawn <hook>: No such file or directory` is
ENOENT on the interpreter, not on the hook, which is present and executable — which
is why the message points at the wrong thing.

The isolated `PATH` now also carries the directory Git for Windows keeps `bash`, `sh`
and `git` in. That directory holds those three and nothing else, so the shell is
granted without putting anything on `PATH` these tests could mistake for a project
tool: `pre-commit` still comes only from the stub, which is exactly what the
no-chain and missing-binary cases assert. Reproduced on a throwaway repo outside the
project before changing anything; adding the directory of `git.exe` alone does not
fix it.

**Where I was wrong.** I reported these as pre-existing at `cd2429d2`, which is true
but incomplete. They are also failing at `4e0ffb46` — the head where this branch's
"49 packages green, 0 red" Windows measurement was taken. So the discrepancy is not
between the tests and the PR description; the measurement itself was wrong.
`internal/ops` was served from the Go build cache on that run, the same way
`internal/commands` was noted as cached that day. A cached package reads exactly like
a measured one, which is the part worth not repeating — every Windows number in the
PR description dates from that run and should be treated as owed a re-measure rather
than as evidence.

**Re-measured at `cfabe1ca`, `-count=1`:** `internal/ops` green in full, 659s, no
failures. That closes the only red I had to report; the other two items in that
section (the dead `SignalProcess`, and `install.sh`'s `*)` branch printing into a
command substitution) stand as reported and are still untouched.

Linux and macOS remain untested here, so the request for review on those still
stands.

jgicquel:
## Rebased, plus three commits the first real Windows run produced

The branch is rebased on `e09aaaa9` and #107 is MERGEABLE again. Two conflicts, both concurrent additions: `.pre-commit-config.yaml` — upstream's `$(go env GOROOT)/bin` is better than the hardcoded path this branch carried, so upstream's version stands and my commit message was rewritten to stop describing a change that no longer exists — and the ADR index, where upstream took 0112, so the Windows ADR is renumbered **0112 → 0113**. The rebase itself moved no file in or out beyond that renumbering, and the branch now stands at 117 files against `e09aaaa9` — the two additions being the new `procscan` files below.

### Scope extension, declared as such

The three commits below answer no finding from either review. They come from running the multi-agent system on Windows for real for the first time — 13 tasks merged across three sprints, native, no WSL — with a launcher window per agent. Two defects came out of it that the suite could not have shown, and I would rather declare them here than have them read as unannounced growth.

**`2a704a93` — `fix(procscan): match the agent image name the way Windows names files`**

This one bears directly on **blocker 4, and on why the closing condition you set was not actually met.**

`IsLizaAgentArgv` trimmed the executable suffix with `strings.TrimSuffix(base, ".exe")` — a case-sensitive comparison applied to a Windows file name. Launching a bare `liza` goes through `PATHEXT`, and MSYS and wezterm hand back `liza.EXE`. So for **every agent started by `launch wezterm mas`, the documented path, identity resolution returned false.**

What that produces, traced to the CLI: `terminateProcess` declines to signal an unverified PID (correct, and deliberately fail-closed), `delete agent` prints the warning that the PID could not be confirmed — and then removes the state row anyway. The agent and its provider stay alive with no row left to find them by. That is the outcome `84c5f561` was written to prevent; the fix is right, it was simply neutralised upstream of itself.

The behavioural test passes because it builds its argv in lowercase. It is green and it proves nothing on this path.

Measured on live agents from a running pipeline: `IsLizaAgentArgv` false → true for the orchestrator and the coder. Agents spawned from the TUI carry a bare `liza agent <role> …` and were never affected — the defect follows the launcher, which is why it hid.

The fix normalises the suffix case on Windows only. On POSIX, `liza.EXE` and `liza.exe` are different files and the case has to stay significant; the two new cases carry `want: runtime.GOOS == "windows"` rather than an unconditional `true`, which would have been the mirror image of the bug — green here, red on Linux.

**`d7ca0c5e` — `fix(procscan): let a host without procfs name its own processes`**

`AgentProcessStatusForPID` established identity by reading `/proc/<pid>/cmdline` and nothing else, so on Windows `AgentProcessLiveMatching` and `AgentProcessMismatched` were both unreachable. Every live agent reported `unknown` — the same answer a PID handed to an unrelated program gives. `get agents` marked all five agents of a healthy pipeline as unverified, and the two callers that act on identity, `clear_stale_review_claims` and `project_cleanup`, exclude *proven*-mismatched PIDs and could never reach that conclusion there.

`procscan.ProcessCommandLine` already answers this; it has served the delete path since termination became identity-checked. It is now routed behind the same entry point, with procfs still authoritative wherever it answers.

One design point worth your eye: **the native source is consulted only when the proc root is the default one.** Around fifteen tests across `ops`, `agent` and `commands` inject a fake procfs precisely to keep the host out of their fixtures; without that guard, a fixture PID that happens to exist on the machine would read as `mismatched` and silently change what those tests assert. A call counter holds the guarantee.

Measured against a live run: `status` went `unknown` → `running`, `get agents` went `✗` → `✓`, on the real agents.

**`fa4a3cd2` — `docs(toolchain): name the second tool that installs on Windows but cannot run`**

The Windows section named `bash-policy` as the platform's only known gap. `scip-python` has the same standing for a different reason: it builds a regular expression from the platform path separator while its module loads, so a backslash leaves the pattern unterminated and the process dies before parsing any argument — `--version` included, which is what `doctor` probes with. A reader could only see an unexplained `doctor` failure on a tool the page says the profile installs. Upstream defect in `@sourcegraph/scip-python`, not worked around here.

### What I have not measured

No full `make test` since the rebase — the numbers in the PR description still stand as retracted, per the correction above. Both fixes were validated by package-scoped runs, a counterproof for each (defect reintroduced, tests red with the reported symptom verbatim), pre-commit on the touched files, and cross-compilation for linux and darwin on amd64 and arm64 plus windows/arm64.

Two failures met along the way, both **pre-existing and proven so**, neither caused by these commits:

- `TestPlanProjectCleanupResolvesRegisteredWorktreeSymlinks`, from `35fda601`, dies in its own setup on Windows: `git worktree list` reports the main repository by its resolved path while the worktree keeps the alias, so the fixture can no longer produce the condition it tests. It has never run on this platform.
- `TestRunSupervisor_HeartbeatMissingAgentStopsSupervisor` fails **in isolation at HEAD with these commits stashed**, `RunSupervisor` returning nil instead of the heartbeat error. That rules out the load sensitivity this package is known for, and it is not in the list of load-sensitive tests I reported earlier. I have not diagnosed it, and I have no reason yet to think it is Windows-specific — flagging it rather than sitting on it.

jgicquel:
## Three more commits from the same real run, declared as scope extension

These do not answer a review finding. They come from running the multi-agent
system on a real project, which put load on paths the test suite reaches only
one at a time. Flagging them as scope extension so they are not read as quiet
growth of an already-large PR.

### 1. A concurrent reader could fail a state write outright (`91d25400`)

State is published by renaming a temp file over `state.yaml`. POSIX allows that
while other processes hold the target open; Windows does not. A single reader
was enough to fail the write, with no retry, and the error surfaced to whichever
agent was mid-transaction.

What I got wrong first, and what the measurement corrected: I assumed this was
`ERROR_SHARING_VIOLATION` and that opening our readers with `FILE_SHARE_DELETE`
would fix it. Neither holds. A probe over all four share modes returns
`ERROR_ACCESS_DENIED` in every case, `FILE_SHARE_DELETE` included — so widening
our own readers' share mode is not a fix.

Nor would locking our readers be enough. `ReadCached` runs without the lock by
design, and the handles that actually span a publish belong to processes no lock
of ours reaches — file-sync clients, indexers, scanners. The retry sits at the
publish point because that is the only place that covers them.

Atomicity is unchanged: the caller already holds the state lock for the whole
attempt and the temp file is unique per call, so a retry republishes exactly the
bytes that were marshalled. `INVARIANTS.md` §5 reads as preserved to me, but
that is the judgement worth checking hardest in review.

The bounded budget is load-bearing rather than cosmetic: Windows reports a
transient collision and a genuinely unwritable file with the same error, so
there is nothing in the error to distinguish them. A real permission failure
spends the budget and then surfaces its own error.

Measured: `os.ReadFile` over a 116 KB state holds the file ~340 µs; a rename in
that window fails. With the retry removed, the covering test fails with the
reported symptom verbatim.

### 2. The zombie scan could not enumerate processes without procfs (`54f87fca`)

`d7ca0c5e` taught per-PID identity to name processes natively. The enumeration
behind `FindZombieAgents` never learned, so `validate` printed
"process scan skipped" instead of a result, and `get agents --zombies` could not
answer.

Measured end to end on a real project, using an inert process built under the
binary's image name and given an agent argv — running an actual agent would have
been the wrong probe:

| | before | after |
|---|---|---|
| impostor running | `WARNING: scan skipped`, then `VALID` | `zombie liza agent process detected: pid 8740 role coder` |
| impostor killed | — | `VALID` |

Scope is decided on the goal ID on this path rather than the working directory,
because Windows does not report another process's cwd without the access rights
`ProcessCommandLine` deliberately avoids needing. Every spawned agent carries
`--goal-id`, so this holds for agents the system started; one launched by hand
without it stays invisible rather than being attributed to the wrong project.

**The procfs path is untouched.** I had first relaxed the shared scope check to
fall back on the goal ID whenever cwd was missing, then reverted it: that would
have changed Linux behaviour, where an unreadable cwd currently excludes a
candidate. Trading accuracy on every other platform for reach on this one is not
a call this PR should make. It is raised in the separate issue instead.

The native path is taken only for the real proc root — the same guard the
per-PID check already carries, held by a call counter — so the fake-procfs
fixtures across `ops`, `agent` and `commands` cannot start seeing the host.

### 3. Two PowerShell 5.1 argument rewrites, documented (`d786fa72`)

Not defects in this codebase: the process never sees the lost argument. Measured
with a program that prints its own argv — an empty value drops out of the
command line and the next flag takes its place (`--reason $empty --agent-id a1`
parses as `--reason=--agent-id`), and double quotes inside a value are stripped.
Spaces, accents, newlines and a leading `--` all survive, so the new
`TROUBLESHOOTING.md` section names the two cases rather than warning about
quoting in general. Both workarounds are measured.

### Not in scope, raised separately

The same run surfaced several behaviours that are **not** platform-specific —
the orchestrator being structurally outside auto-repair, the 4096-byte reason
cap with no `--reason-file`, `mandatory-docs: []` against the base prompt's
promise that later agents read `lessons/`, and the POSIX cwd question above.
They are in a separate issue as questions rather than patches, because several
of them may well be deliberate trade-offs.

### Validation

`gofmt`, `go vet`, `staticcheck` and the full pre-commit set clean on the
touched files. `internal/db`, `internal/procscan` and `internal/testguard` green
with `-count=1`; `internal/commands` green on the tests that assert on this
behaviour. Cross-compiles clean for linux/amd64, darwin/amd64 and windows/arm64.
Each fix has its counter-proof: reintroduce the defect, watch the test fail with
the real symptom's message.

`internal/agent` (255 s) and `internal/ops` (820 s) each end with one failure.
Both are red on pristine `upstream/main` (`e09aaaa9`) under Windows too — checked
in a detached worktree, so this is measured rather than assumed:

- `TestPlanProjectCleanupResolvesRegisteredWorktreeSymlinks` fails identically
  upstream and here, inside its own fixture: `git worktree list` reports the main
  repository by its resolved path in forward slashes and the linked worktree
  through the alias in backslashes, so the setup's own precondition never holds.
  It reaches `PlanProjectCleanup` unchanged.

- `TestRunSupervisor_NonzeroAgentErrorBlocksBeforeAutoRepairCanRespawn` is worth
  a closer look, because **this branch improves it without fixing it**. Upstream
  it fails with `Run calls = 0` and `agent coder-1 registered process PID … is
  not running` — the claim never succeeds, since identity cannot be established
  without procfs. Here it reaches `Run calls = 1`: the claim works, the agent
  runs and crashes as designed. It stops one short because the test's budget is
  `10*time.Second` (`supervisor_test.go:1532`) while on this machine the claim
  alone costs ~7 s in git worktree setup, and the supervisor then waits
  `delay_seconds=5` before the second run — 12 s needed against 10 available.
  Where git is fast that margin exists, which is presumably why it passes on CI.
  I have not touched it: raising a timeout in someone else's test is not this
  PR's business, and you may prefer a different fix. Flagging it in case the
  budget is worth revisiting independently of Windows.

jgicquel:
## Rebased onto `4a471c76`

74 commits, 123 files, +5013/−657. The touched file set is identical before and
after the rebase — no file gained or lost — and the helper counters this branch
tracks (`RequireSymlinkCapability`, `WriteShellStub`, `ResolveBashForScripts`,
`paths.UserHomeDir`, `PreventRemoval`, `stubrelay`) are unchanged.

Three of the four new upstream commits close points I had raised in #117.
`33a84b78`, `d3e87143`, `ead61987` and `4a471c76` all run green on Windows here:
`internal/statehygiene` 3.6 s, `internal/prompts` 6.5 s, `internal/embedded`
327 s, `internal/commands` 1067 s, `cmd/liza` 463 s, `internal/procscan` 2.6 s,
`internal/db` 11.6 s, `internal/agent` 405 s — all `-count=1`.

`internal/ops` (1397 s) ends with one failure,
`TestPlanProjectCleanupResolvesRegisteredWorktreeSymlinks`, which is red on
pristine upstream under Windows too — I checked it in a detached worktree, and
nothing upstream touched that test between `e09aaaa9` and `4a471c76`. It dies in
its own fixture: `git worktree list` reports the main repository by its resolved
path in forward slashes and the linked worktree through the alias in
backslashes, so the setup's precondition never holds and `PlanProjectCleanup`
itself is never reached.

One correction to what I wrote yesterday about the other failure I had reported,
`TestRunSupervisor_NonzeroAgentErrorBlocksBeforeAutoRepairCanRespawn`: it now
passes as part of the full package run. It still fails in isolation, three times
out of three. That is the reverse of the usual pattern, and it fits the cause I
described — the test's `10*time.Second` budget against a claim that costs ~7 s
here plus a 5 s crash-restart delay. Run alone, the first claim pays for git
repository, worktree and index setup cold; run after the rest of the package,
those caches are warm and the claim fits. Marginal rather than broken, and the
margin is a property of the machine, not of this branch.

### One conflict, and the design call inside it

`ead61987` and this branch both rewrote `internal/procscan/liza_agents.go`.
Upstream's structure stands: `ZombieScanResult`, the three-way `classifyScope`,
and the procfs loop are taken as they are, with only `newZombieProcess`
substituted for the inline literal and `defaultProcRoot` for the `/proc` string.
The Windows delta is confined to the non-procfs fallback.

That fallback needed a decision I could not take from the code alone. On a host
with no procfs the working directory is not unreadable, it is absent by
construction — Windows does not report another process's cwd without the access
rights `ProcessCommandLine` deliberately avoids needing. Classifying every
candidate as unknown scope on that basis would report the entire agent pool as
unresolved on every single scan, which is the noise this path exists to remove.

So the goal ID each candidate carries in its own argv stands in for the cwd, and
your three outcomes stay meaningful rather than collapsing to one:

| candidate | outcome |
|---|---|
| goal ID matches the scan's | `Zombies` |
| goal ID belongs to another run | dropped, as foreign |
| no goal ID in argv | `UnknownScope`, reason `cwd_unreadable` |

The procfs path is untouched by this — its cwd evidence is stronger, and
weakening it to match would trade accuracy everywhere else for reach here. If
you would rather every Windows candidate land in `UnknownScope` regardless, say
so and I will change it; the structure supports both and this is your call more
than mine.

### A silent breakage the rebase introduced

`cmd/liza/main_test.go` stopped compiling: upstream added `"strings"` for the
flag-consumed-as-reason test, this branch had removed it when the platform test
that used it was replaced (upstream asserts Windows is rejected; this branch
asserts all three platforms are accepted). Git merged both hunks without
conflict and kept the removal.

Worth flagging because of how it hid: `go build ./...` stayed green throughout,
since it does not compile test files. Only `go vet ./...` caught it. Fixed in
its own commit rather than folded into the conflict resolution.

### Not folded in: the remedy your CLI guard suggests

`validateReasonFlag` catches the shifted argument correctly. Its message
recommends `--reason "$reason"`, which fixes it in a POSIX shell and does
nothing in Windows PowerShell 5.1 — measured, quoting does not stop an empty
argument from being dropped before the process starts. Details and the form that
does work are in #117; I have not touched `main.go` here, since you wrote it
hours ago and may prefer your own wording.

tangi-vass:
Thanks — decisions on the remaining non-review questions:

1. Keep the current goal-ID fallback for scope classification on hosts without procfs.
2. Please update `validateReasonFlag` guidance to include the PowerShell-safe attached form, `--reason="$reason"`.
3. We’ll merge #106 first. Once it lands, please rebase #107 onto `main` so the duplicated test commits drop out of this PR.
4. `LIZA_TOOL_*` should be white-labeled. I consider these variables part of the white-label boundary rather than an internal exception. Please update the variables introduced or touched here and keep the family consistent.

jgicquel:
Thanks — 2 and 4 are done and pushed (`c9b92134`, `4b19c901`). 1 needs nothing.
3 is waiting on #106.

### 2 — the attached form

Before writing your wording into an error message I measured the exact spelling
you gave, since mine in #117 was `"--reason=$reason"` and yours is
`--reason="$reason"`. They are equivalent: identical argv, byte for byte.

| invocation | argv the process receives |
|---|---|
| `--reason $reason` | `--reason`, `--agent-id`, `a1` |
| `--reason "$reason"` | `--reason`, `--agent-id`, `a1` |
| `--reason="$reason"` | `--reason=`, `--agent-id`, `a1` |

The message now carries both forms and says why the second exists, so a reader
on a POSIX shell does not wonder what the extra one is for.

`support-docs/TROUBLESHOOTING.md` documented this rewrite already but spelled
the workaround the other way. Two spellings of one remedy in a single product
have no reason to differ, so it now matches the message. Its example also used
`$reason` before assigning it; the two lines are in order now.

### 4 — the whole family, not only ours

I took "keep the family consistent" to mean the family rather than the four this
branch introduced or touched, since three hardcoded names beside four resolved
ones is the inconsistency worth avoiding. Seven variables now go through
`brand.EnvName`, via one `toolEnvName` helper:

`TOOL_ARCHIVE_URL`, `TOOL_BINARY`, `TOOL_INSTALL_DIR`, `TOOL_INSTALL_URL`,
`TOOL_SOURCE_REPO`, `TOOL_SOURCE_PACKAGE`, `TOOL_PACKAGE`.

`INSTALL_DIR` and `Tool.InstallDirEnv` stay as they are — those names are
dictated by third-party installers, not by us. No legacy alias either: unlike
the variables `envgate` reads from the environment, these are written and read
back inside the same function, so nothing outside the package can be relying on
the old names.

Verified against a non-default-brand build rather than only against the tests.
`toolchain install --dry-run` from a binary built with `EnvPrefix=ACME_AGENT`:

```
[PLANNED] bash-policy :: bash -c curl -fsSL "$ACME_AGENT_TOOL_INSTALL_URL" | bash
[PLANNED] rtk :: powershell ... Invoke-WebRequest -Uri $env:ACME_AGENT_TOOL_ARCHIVE_URL ...
```

No `LIZA` literal anywhere in that output. The new test pins all seven names
spelled out rather than derived, so a `toolEnvName` that stopped consulting the
brand would fail it instead of agreeing with itself, and it checks that the
scripts and the environment agree — a variable renamed on one side only would
leave the install broken while every name still looked right.

### 3

Watching #106. I will rebase onto main as soon as it lands and confirm the
duplicated test commits drop out.

One thing I noticed while checking this PR's state: `gh pr checks` reports no
checks on the branch, so nothing is exercising Linux or macOS here. The reviewer
asked about that earlier and I answered that I could not test those platforms
locally — worth saying plainly that CI is not covering them either.

tangi-vass:
Independent review at `4b19c901`. Standard depth, deepened on the paths that have had no reviewer pass at all: everything after `cd2429d2` — the six declared scope extensions, the two rebases, and the two commits answering the maintainer decisions.

## First: the Linux measurement you asked for, twice

`make test` (`go test -race -timeout 30m`) on linux/amd64, in a detached worktree pinned to `4b19c901`:

```
47 packages ok · 2 packages FAIL · 0 data races
```

Three of the five failing tests are pre-existing. I checked them at your own merge base `4a471c76` with none of this branch applied, and they fail there identically (`invalid pipeline config: field task-slug not found in type pipeline.RolePairDef`). Not yours.

**The other two are yours, and they are the blocker below.**

## Reconciliation

Verified against source at this head, not against thread state. All nine findings from the two reviews at `cd2429d2`: **RESOLVED**. Specifically re-derived rather than taken on report — `install.sh` refuses on stderr before any download; `gofmt -l .` is empty tree-wide; `project_dir` is normalised at `enforce-init.sh:133`; `windowsLaunchShell` probes the three Git for Windows roots ahead of `LookPath`; `ProcessAlive` gates on `GetExitCodeProcess`; `install.ps1` help is brand-neutral; `packageInstallCommand` continues past a manager that names no package. Maintainer decisions 1, 2 and 4 are likewise done — the seven `TOOL_*` names all route through `toolEnvName`.

Decision 3 is not. More on that below.

---

## [blocker] `internal/procscan` — two tests this branch added fail on Linux

```
--- FAIL: TestFindZombieAgents_NativeScanMatchesOnGoalID
        liza_agents_test.go:238: zombie count = 0, want 1
--- FAIL: TestAgentProcessStatusForPID_NativeIdentityMatches
        process_status_test.go:39: DisplayStatus() = "mismatched", want "running"
```

Deterministic, 3/3 runs, `-count=3`.

**Root cause.** Both stub the native probe with a Windows-shaped `argv[0]`:

```go
liza_agents_test.go:228:   []string{`C:\bin\liza.EXE`, "agent", "coder", "--cli", "codex", "--goal-id", "goal-1"}
process_status_test.go:34: []string{`C:\bin\liza.EXE`, "agent", "coder", "--agent-id", "coder-1"}
```

`IsLizaAgentArgv` takes `filepath.Base(argv[0])`. On Linux the backslash is not a separator, so nothing is stripped, and `trimExecutableSuffix` then takes its case-sensitive branch — which does not match `.EXE`. Measured:

```
filepath.Base("C:\\bin\\liza.EXE") = "C:\\bin\\liza.EXE"
TrimSuffix(base, ".exe")           = "C:\\bin\\liza.EXE"
== "liza" ?  false
```

So identity resolves false, and both assertions invert. The two sibling tests in the same files pass only because they assert the *negative* outcome and reach it by the wrong route.

**Why it matters beyond the red.** This is the mirror image of the bug you fixed in `2a704a93` — and you named it there exactly: *"the two new cases carry `want: runtime.GOOS == "windows"` rather than an unconditional `true`, which would have been the mirror image of the bug — green here, red on Linux."* That discipline was applied in the suffix cases and not in the native-path tests added two days later.

**Why it was invisible.** `git log -S 'C:\bin\liza.EXE'` dates them to `67adbfb1` (08-14) and `54594e64` (08-15). The only Linux sweep this branch has ever had was mine at `cd2429d2` on 08-10. Both commits land in the window you flagged as unmeasured, and nothing has measured it since.

**Likelihood:** every CI run and every contributor's `make test` on Linux or macOS.

**Closure condition:** `make test` green on linux/amd64 at the reviewed head, with the two tests either exercising a platform-appropriate `argv[0]` or gating their expectation on `runtime.GOOS` as `TestIsLizaAgentArgv` already does.

## [blocker] The branch does not merge, and decision 3 is outstanding

`mergeStateStatus` is `DIRTY`. Nine files conflict against `main`, and they are not all test files:

```
Makefile
internal/ops/wt_create.go
internal/testhelpers/setup.go
specs/architecture/ADR/README.md
cmd/liza/brand_smoke_test.go
internal/agent/checkpoint_summary_test.go
internal/db/blackboard_test.go
internal/ops/{await_verdict,claim_task,delete_task}_test.go
internal/procscan/liza_agents_test.go
```

#106 landed as `e129bc3c`, and `main` has since moved to `5c4056fa`. The rebase you were waiting on is unblocked.

One thing the rebase will force, worth knowing before you start: **ADR-0113 collides.** `main` already carries a merged `0113-sliced-integration-analysis-and-final-closure.md`, and upstream is now at 0117. This branch's `0113-native-windows-support.md` has a different filename, so git will merge both without a word and leave two decisions numbered 0113. Next free number is 0118. This is the third renumber (0112 → 0113 → 0118); the ADR index conflict is the only thing that will flag it, so it is worth doing deliberately rather than during conflict resolution.

---

## [concern] `.github/workflows/ci.yml:51` — the Windows job cannot fit the budget it was given

ADR-0113 makes this job load-bearing: *"CI runs on windows-latest. Without it the rest is a claim rather than a property, and every future change can regress it silently."*

The job sets `go test -timeout 15m ./...` and `timeout-minutes: 45`. Against your own published measurements:

| package | your measurement | vs. 15m per-package |
|---|---|---|
| `internal/ops` | 17m55s (`-race`), later 23m17s (`-count=1`) | over |
| `internal/commands` | 10m12s (`-race`), later 17m47s (`-count=1`) | over on the second |
| ops + commands + `cmd/liza` + agent + embedded | ~61 min | over the 45m job cap |

`-timeout` is per test binary, not per run, so `internal/ops` alone exceeds it. `make test` carries `-timeout 30m` with a comment citing those same numbers; the CI job carries half that. Hosted `windows-latest` runners are not faster than your machine.

The failure mode is the one you documented in the PR description: *"the tests after the one running at that moment are never reported, so a truncated measurement reads like a handful of failures rather than an incomplete run."* A first-ever run that truncates will read as a handful of Windows defects.

Compounding it: this job has never executed. The workflow run is still gated, and `license/cla` is the only check present at this head — and the only *required* one under the `protect-main` ruleset. So nothing will ever turn this red job into a signal anyone has to act on.

**Closure condition:** the Windows job's per-package and job budgets are consistent with the branch's own measurements, or the job is observed completing once at the reviewed head.

## [concern] `internal/db/publish_state_windows.go:59` — the retry does not cover a scanner holding the temp file

Flagging this because you asked for §5 to be checked hardest. The atomicity argument holds — the lock spans the attempt, the temp file is unique per call, and one rename still either happens or does not. This is about the error set, not atomicity.

`publishState` retries `ERROR_ACCESS_DENIED` (5) only. That is the code for a blocked *destination*, which is what you measured. The rename has a second operand: the temp file, freshly created in the same directory and closed microseconds earlier. A scanner that opens *it* blocks the rename as its source, and Windows reports that as `ERROR_SHARING_VIOLATION` (32) — which falls straight through `errors.Is(err, errAccessDenied)` and returns with no retry at all.

The actors are the ones your own comment names: *"file-sync clients, indexers and scanners walking the project tree."* Real-time scanning of a just-written file is exactly what those do, and it is guaranteed on the managed machines this PR targets.

I have no Windows host — this is reasoned from platform semantics, not measured, which is why it is a concern and not a blocker. You have the host and a probe harness already written.

**Closure condition:** either `publishState` also retries `ERROR_SHARING_VIOLATION`, or a measurement records why a handle on the temp file cannot block the rename.

## [concern] `Makefile:115` — `SHELL := bash` resolves the shell by PATH

```make
SHELL := bash
.SHELLFLAGS := -c
```

This is the trap `windowsLaunchShell` exists to avoid, in the one place that still walks into it. `C:\Windows\System32\bash.exe` is the WSL launcher, the machine PATH is searched first, and a user-level Git entry cannot win — your own `TROUBLESHOOTING.md:490` says so and prescribes an elevated fix.

The reason I do not think this is out of scope: the block just above it explicitly anticipates a non-Git-Bash caller — *"HOME is not a Windows variable... From PowerShell or cmd it is empty"* — and falls back to `USERPROFILE` for exactly that case. So `make install` from PowerShell is a supported invocation, and on that invocation `SHELL := bash` picks whatever PATH offers. The recipes then run under WSL against a `C:/...` install directory.

The candidate list is already written twice in this branch (`windowsLaunchShell`, `ResolveBashForScripts`).

**Closure condition:** the Windows `SHELL` names a Git for Windows `bash.exe` by path, or the Makefile documents that a Git Bash invocation is required and the `USERPROFILE` fallback is dead code.

---

## [vestigial] `internal/ops/delete_agent.go:65` — `SignalProcess` outlived the design it belonged to

You reported this twice and got no answer, so: it is still here, and this PR is what made it vestigial. `terminateProcess` now goes through `signalTree`/`killTree` with identity verification, a grace period, and `IdentityUnverified` reporting. `SignalProcess` is the last SIGTERM path, has no production caller, and `proc.Signal(syscall.SIGTERM)` returns "not supported" on Windows so it can only ever return false there.

Its doc comment is now wrong on both counts it makes — *"sends SIGTERM"* and *"Verifies the process is an agent via `/proc/<pid>/cmdline`"* — and it describes the exact defect class four commits of this branch were spent closing. That is the part that will mislead someone.

Not blocking: nothing calls it, so it worsens neither net value nor risk. Delete it, or say why it stays.

## [suggestion] `install.sh:127,137` — the sibling refusals still print into the command substitution

Also self-reported and unanswered. The comment you added at line 114 states the invariant — *"this function's stdout is the caller's command substitution, which would swallow them"* — directly above three branches that violate it: unsupported OS at 127, unsupported architecture at 137. A user on FreeBSD or riscv64 gets a silent `exit 1`.

Three `>&2`. Pre-existing and outside every finding, so this routes to follow-up rather than to this round — but the new comment makes the omission read as deliberate.

## [suggestion] `internal/ops/delete_agent_process_windows.go:98` — the tree walk has no visited set

Two things ride on the same three lines.

The BFS over `childrenByParent` guards only `child == root`. Stale parent pointers plus PID recycling can produce a cycle (A's recorded parent is B, B's is A), and the creation-time comparison breaks it only when *both* times are readable. When one is not, `order` grows without bound and `liza delete agent` hangs.

That same unreadable case is also kept rather than dropped — `processCreationTime` returns `(0, false)` and the candidate stays in a list every member of which is about to be terminated. For a kill operation the safe default is the other one. In practice a process you cannot open for `PROCESS_QUERY_LIMITED_INFORMATION` is one you cannot open for `PROCESS_TERMINATE` either, so today this costs an error rather than a wrong kill — but that is an accident of the two access rights, not a property the code states.

A `map[uint32]bool` closes both.

## [suggestion] `internal/embedded/hooks/enforce-init.sh:102` — the comment omits the one metacharacter that stops being rejected

The comment reassures that *"`;`, `&`, `|`, `<`, `>`, `` ` `` and `$(` remain rejected"* after the separator rewrite. The reject regex at line 253 is `[;&|<>`\\]|[$][(]` — it also rejects `\`, and on Windows the rewrite at line 104 makes that arm unreachable.

I could not construct a bypass, and I do not think one exists: the rewrite cannot *create* any of the other rejected characters, and an escaped `;` is still a `;`. But the comment enumerates a list and leaves out the one entry the change actually affects, which is how a future audit concludes the guard is unchanged.

---

## [appraisal]

Three, briefly.

`renameOverBinary` — displace, install, sweep, and restore the previous binary if the second rename fails. Self-replacing binaries are where updaters lose people's tools, and the failure path is the one usually left out.

The `defaultProcRoot` guard on the native probe. Roughly fifteen fixtures across `ops`, `agent` and `commands` inject a fake procfs to keep the host out; consulting the native source only for the real proc root is what stops a fixture PID that happens to exist from silently changing what those tests assert. That is a second-order consequence found before it bit.

And `d786fa72` — measuring PowerShell 5.1's argument rewrites with a program that prints its own argv, then documenting the two cases that actually lose content instead of a general warning about quoting. Then re-measuring the maintainer's exact spelling against your own before writing it into an error message. Both are the expensive move and the right one.

---

## Review summary

**Review:** Standard (deepened on the unreviewed post-`cd2429d2` surface) — **Request Changes**

- **Blockers:** 2
- **Concerns:** 3
- **Suggestions:** 3 (+1 vestigial)

**Overall:** The engineering here is consistently better than the review it has received — every defect is reached by a measurement, every trade-off is stated in the source rather than the PR, and the corrective rounds fixed the findings rather than the symptoms. What is failing is not the work but its verification loop: the only Linux sweep predates a third of the branch, the Windows CI job that ADR-0113 makes load-bearing has never executed, and two tests written for Windows semantics have been red on Linux for sixteen days.

**Approach:** Round 1 for this reviewer, so it is in scope. Sound, and I weighed the alternative — split the test-harness work (`stubrelay`, `perm`, `removal`, `filemode`) into its own PR ahead of the production fixes. It loses: that block is the least risky part of the change and splitting it would not shrink what actually needs measuring.

**Absence:** Baseline taken from the PR description's Change map, its Not-in-scope list, ADR-0113, and the maintainer decisions of 08-16. Present and checked. What is missing is a current Linux measurement — which is the blocker, not a separate gap.

**Net value:** *As submitted* — negative: it does not merge, and it makes `make test` red on the platform the project actually develops on. */ With the findings resolved* — clearly positive: nine platform-independent defects fixed, a platform restored under stated conditions, and capability probing rather than `GOOS` skips, which is the difference between a port and a port you can trust.

**Blast radius:** High — install and distribution, agent process lifecycle, state publication, the initialization gate, path canonicalisation.

**Confidence:** High on the Linux blocker and the CI budget — both measured or derived from your own published numbers. Medium on the three Windows-semantics concerns: reasoned from platform behaviour, no Windows host here, and each names what would settle it.

**Sources:** full base-to-head diff (123 files); exact-head source in a detached worktree at `4b19c901` for `install.sh`, `install.ps1`, `Makefile`, `.github/workflows/ci.yml`, `.goreleaser.yaml`, `internal/db/publish_state_*`, `internal/procscan/*`, `internal/ops/delete_agent*`, `internal/updater/updater.go`, `internal/toolchain/install.go`, `internal/embedded/hooks/*`, `internal/pairingindex/hooks.go`, `cmd/liza/cmd_launch.go`, ADR-0113; `make test` at head and a targeted baseline run at `4a471c76`; `gofmt -l`, `go vet`, `go build`; `git merge-tree` against `main`; the ruleset and check state; all ten PR comments; `GUARDRAILS.md`.

**Next step:** rebase onto `main` (renumbering the ADR to 0118), fix the two Linux failures, then one focused re-review — the three concerns can be answered in the same round or contested with a measurement.

*On the two you flagged and I did not fix:* `TestPlanProjectCleanupResolvesRegisteredWorktreeSymlinks` and the `TestRunSupervisor_...` budget are both yours-were-right calls. Neither is this PR's to carry; both deserve their own issue, and the second is a real 10s-budget fragility independent of Windows.

tangi-vass:
Review: Deep — Request Changes

Blockers: 2 new
Concerns: None new
Suggestions: None new

Overall: The previous review chain’s findings are resolved, but two newly added native-Windows paths still have deterministic failure behavior: an interrupted upgrade can remove the working executable, and PowerShell activation is generated but not presented to the user.

Approach: Round 2; the native-Windows/Git-for-Windows approach is settled and was not relitigated.

Absence: Failure-path coverage is missing for replacing an existing installation through `install.ps1`, and CLI coverage does not exercise Windows PowerShell activation output.

Sweep: Performed because the design expanded after the previous round. No additional P0-P2 issue was found beyond the two below. The concurrent [independent review](https://github.com/liza-mas/liza/pull/107#issuecomment-5477455930) already covers the red `procscan` tests, merge conflicts/ADR collision, CI budget, state retry, Makefile shell selection, and remaining nonblocking observations; none are duplicated here.

Net value: As submitted — negative, because upgrading can leave the CLI absent and the native PowerShell setup path emits an unusable activation command. / With these findings and the independent review’s blockers resolved — clearly positive: native Windows support is restored while preserving the existing platform contract and the cross-platform fixes exposed by the port.

Blast Radius: High — installation, distribution, shell activation, process lifecycle, and public platform support.

Confidence: High for these two findings; both are direct source-level failure paths. Overall Windows confidence remains medium because no Windows runtime host or completed Windows CI run was available.

Sources: PR description and issue #104; complete 123-file base-to-head change set; exact-head source at `4b19c901`; prior comments and maintainer decisions; project invariants, guardrails, architecture review, and ADR registry; exact-range pre-commit; `make test`; Windows amd64/arm64 cross-builds; refreshed comments, reviews, mergeability, ruleset, and check status.

Next step: rebase and resolve the independently reported blockers, fix both findings below, obtain completed Linux and Windows validation, then request a focused re-review.

Round 2 — remaining blockers/concerns 7→2 in this chain, files 107→123

Reconciliation: All actionable findings from the prior `pr107-cd2429d-round1` review are RESOLVED at this head: `install.sh` now refuses Windows before downloading; Git for Windows is preferred by `launchShell`; process liveness checks the exit code; unknown package-manager entries continue; formatting is clean; tool environment names use the brand prefix; `CLAUDE_PROJECT_DIR` is normalized; native deletion verifies identity and reports unverifiable live processes; and `install.ps1` help respects the white-label boundary.

## [blocker] `install.ps1:117-126` — a failed replacement can remove the working installation

When `$target` exists, the script renames it to `$target.old`, immediately attempts to delete that displaced copy, and then moves the downloaded binary into place. If the final `Move-Item` fails—because of antivirus scanning, permissions, a filesystem race, or another write failure—the `finally` block only removes the temporary directory. There is no catch or rollback restoring the previous executable.

Why it matters: An upgrade failure can turn a usable installation into no installation at all. This is the failure path the self-updater’s displacement logic already protects against.

Likelihood: Any upgrade where the existing target is present and the final move fails after displacement. Managed Windows machines with active endpoint scanning are part of the stated target environment.

Closure condition: Preserve and restore the displaced executable whenever installation of the replacement fails, or provide an equivalent atomic/rollback-safe replacement sequence. Add regression coverage that forces the final move to fail and verifies the original executable remains runnable.

## [blocker] `cmd/liza/cmd_toolchain.go:167-169` — native PowerShell users receive the POSIX activation command

On Windows, `internal/toolchain/configure.go:99-123` deliberately writes `env.ps1` and returns its path through `PowerShellEnvPath`. The human-readable CLI output ignores that field and always prints only:

```text
Wrote .../env.sh
Run: source .../env.sh
```

`source` is not a PowerShell command and `env.sh` is not PowerShell syntax. Without `--write-shell-profile`, the generated `env.ps1` is left unwired and the command gives the user no usable activation instruction. This also conflicts with the documented contract that `configure` prints the exact command for the current shell.

Why it matters: Following the command’s own output in the native Windows shell does not activate the selected toolchain or feature gates, so subsequent initialization and agent commands can behave as though configuration never happened.

Likelihood: Every `toolchain configure` invocation from PowerShell without `--write-shell-profile`.

Closure condition: On Windows, human-readable output reports the generated `env.ps1` and prints a valid PowerShell dot-source command, while retaining the `env.sh` instruction for Git Bash. Cover the Windows CLI-output behavior with a regression test.

Validation:
- ✅ `git diff --check`
- ✅ Exact-range pre-commit, with `GOFLAGS=-buildvcs=false` to avoid linked-worktree VCS-root misidentification
- ❌ `make test`: the two new cross-platform `internal/procscan` tests fail; already reported by the concurrent independent review
- ✅ The three apparent `cmd/liza` failures pass with an isolated process `HOME`; they were caused by the host’s newer global pipeline schema and are not attributed to this PR
- ✅ Windows `amd64` and `arm64` cross-builds
- ⚠️ No Windows runtime host; GitHub currently exposes only the successful required `license/cla` check

tangi-vass:
Superseded by #147

PR #74, opened 2026-05-19
feat(cli): add pi (pi-coding-agent) as a supported CLI backend

ajanderson1:
## Summary

Adds [Pi](https://github.com/badlogic/pi-mono) (a.k.a. `pi-coding-agent`) to Liza's `validCLIs` list, joining `claude`, `codex`, `gemini`, `mistral`, `kimi`. Spawned via `pi -p [--mode json]`, single-shot, prompt piped on stdin — the same shape as the existing adapters.

**Why Pi.** Pi is provider-agnostic by design — one CLI fronts Anthropic, OpenAI, Google, DeepSeek, Bedrock, Groq, OpenRouter, GLM, Kimi, Mistral, and ~15 others. Wiring it in lets Liza users pick any of those providers per role without a per-CLI Liza integration each time.

## Changes

| File | Change |
|------|--------|
| `internal/agent/cli.go` | `validCLIs` += `"pi"` |
| `internal/agent/supervisor.go` | `case "pi":` in `Execute` switch — `pi -p` (stdin-piped prompt), `--mode json` when output capture is enabled |
| `internal/commands/init.go` | `"pi" -> "AGENTS.md"` symlink (pi auto-discovers `AGENTS.md`); no-op case in classifier — no PreToolUse hook setup |
| `cmd/liza/cmd_init.go` | `--pi` flag registered; `"pi"` added to `agentFlagNames` |

No new packages, no new dependencies. `ExecuteInteractive`'s `default:` branch (`exec.Command(actualCLI)`) already handles `pi` interactive mode.

## Verified end-to-end on macOS

```bash
liza init --pi --default-cli pi "Build tiny todo CLI" --spec specs/vision.md
# AGENTS.md → ~/.liza/CORE.md, state.yaml.config.default_cli: pi

liza agent orchestrator --cli pi
# orchestrator-1 spawned, read contract via pi `read` tool calls,
# wrote epic-1 via `liza add-task` (model: opencode-go/glm-5)

liza tui
# epic-1 DRAFT_EPIC_PLAN visible; orchestrator-1 IDLE; hotkeys reachable
```

## Test plan

- [x] `make build` clean
- [x] `make test` passes on touched packages (`internal/agent`, `internal/commands`, `cmd/liza`)
- [x] `liza init --pi` creates `AGENTS.md` symlink to `~/.liza/CORE.md`
- [x] `liza init --default-cli pi` is accepted and persisted to `state.yaml`
[...]

PR #162, opened 2026-09-22
feat(providers): add pi backend with init-gate enforcement and model-selection support

LivioGama:
## Summary

Adds [pi](https://github.com/badlogic/pi-mono) (pi-coding-agent) as a supported CLI backend, built on the current provider-catalog architecture, plus the enforcement and model-selection pieces needed to run it for real.

Supersedes #74 — that PR targets an architecture that main has since replaced (hardcoded `validCLIs`/supervisor switch no longer exist) and now conflicts with main in all four of its files. This branch delivers the same intent in the catalog model, credits and preserves @ajanderson1's original direction (spawn shape, `AGENTS.md` contract, `--pi` flag), and implements the follow-ups that PR listed as out of scope (native tool-call gating, model selection).

## Changes

| Area | Change |
|------|--------|
| `provider-catalog.yaml` + `internal/providers/embedded.go` | `pi` provider entry: stdin prompt transport, `pi -p` (+ `--mode json` when logging), `AGENTS.md` contract with preferred-global placement, `pi.env` API-key injection, `pi_extension` activation asset |
| `internal/embedded/pi-extensions/liza-init-gate.ts` | pi counterpart of `enforce-init.sh`: blocks mutating tools until the behavioral contract docs and `GUARDRAILS.md` are read; absent docs are skipped (no deadlock) and read detection is symlink-safe (`realpathSync`) |
| `internal/embedded/embedded.go` | `WritePiInitGate` (branded global deployment, managed-header policy); `pi_extension` activation asset writer |
| `internal/agent/launch_resolver.go` | new `{{globalDir}}` template variable so catalog `run_args` can reference branded global assets |
| `internal/commands/init.go` | pairing/full-init deployment of the gate; `CheckContractConfigured` backfills the embedded catalog so new providers resolve before the fetched catalog catches up |
| `internal/commands/setup.go` | `liza setup --pi` deploys the gate alongside skill symlinks |
| `cmd/liza/cmd_agent.go` + `internal/commands/providers.go` | spawn-path self-heal: `EnsureProviderSpawnAssets` redeploys the gate if missing (pi refuses to start on a missing `-e` path); respects the managed-header policy |
| `cmd/liza/cmd_init.go` | `--pi` flags for init and setup |
| `provider-catalog.yaml` (codex) | `codex.env` env-file so provider API keys (e.g. `ZAI_API_KEY` for z.ai GLM) reach codex spawns |
| `docs/provider-models.md` | provider model-selection guide: pi setup, Devin SWE-2 Max selection, GLM 5.3 Flash via codex |

## Model selection

No new config surface: `agent_tools.<id>.run_args` overrides in `state.yaml` select models, e.g. Devin → SWE-2 Max:

```yaml
agent_tools:
  devin:
    run_args: [--permission-mode, dangerous, --model, swe-2-max, -p, "{{prompt}}"]
```

and GLM 5.3 Flash through Codex via a `model_providers.zai` block with `env_key` (see docs).

## Verified end-to-end on macOS (devin 3000.11.1, pi 0.84.4, codex 0.155.1)

- `liza init --pi` (pairing and full init): global contract symlink + gate deployed; `default_cli: pi` persisted
- `liza agent orchestrator --cli pi`: real pi subprocess, planning turn streamed
- Init gate live: bash blocked until contract docs + `GUARDRAILS.md` read; absent-doc workspaces no longer deadlock; unlocks after reads
- `devin --permission-mode dangerous --model swe-2-max -p` → `SWE2-MAX-OK`; Liza-spawned Devin sessions confirmed
- `codex --profile zai-glm exec` with `ZAI_API_KEY` → GLM-5.3-Flash responses
- Subprocess tests assert spawn argv/stdin, the `--model swe-2-max` override, and codex env-file propagation

## Test plan

- [x] `go build ./...` clean; `go vet` clean on touched packages
- [x] `internal/providers`, `internal/embedded`, `internal/commands`, `internal/agent`, `cmd/liza` all pass (`-count=1`)
- [x] Only pre-existing environment failures remain (`worktreeexclude`/`functionalclusters`/ops excludesFile tests fail identically on pristine main when a global `core.excludesFile` is configured)


tangi-vass:
**Review: Standard (Deep on the catalog + gate) — Request Changes (comment)** · head `16cd2e0f`

Blockers: 2 · Concerns: 2 · Suggestions: 4

**Overall:** The pi backend is built the right way — catalog entry plus the existing env-file, activation-asset and template-var mechanisms, no new config surface. Two things block: publishing a catalog field that older binaries cannot parse, and a raw default-brand filename in a generated artifact. Both are cheap now and expensive after this merges.

---

## `[blocker]` `provider-catalog.yaml:293` — `pi_extension` makes the published catalog unparseable by every already-installed binary

`ParseCatalog` decodes with `dec.KnownFields(true)` (`internal/providers/catalog.go:215`), and `DefaultCatalogURL` is `raw.githubusercontent.com/liza-mas/liza/main/provider-catalog.yaml` (`catalog.go:34`) — so merging publishes this file to every install on its next TTL refresh.

Verified by running the pre-merge parser (current `main`) against this branch's catalog:

```
yaml: unmarshal errors:
  line 293: field pi_extension not found in type providers.ActivationAssets
```

**Why it matters:** `Load` swallows the parse failure and falls through to the stale cache, then to `EmbeddedCatalog()` (`catalog.go:611-626`), with no warning. The failure repeats on every refresh, so *every future catalog change* — not just `pi` — becomes invisible to every pre-merge client until they upgrade the binary. That is the one thing the remote catalog exists to avoid.

**Likelihood:** Certain on merge, for the whole installed base. This is the first field added to `ActivationAssets` since the registry shipped — `mistral_prompt_config`, `bash_policy_claude` and `bash_policy_codex` all date to `61adbb10f`, so this path has never been exercised.

**Suggestion:** keep `pi` in the embedded catalog only for one release. It works fully from there: `resolveCatalogProviders` (`internal/commands/providers.go:38-44`) and `agentToolsFromCatalogs` (`internal/agent/launch_resolver.go:88-95`) already backfill embedded built-ins, and `TestRepositoryCatalogAddsRemoteProviders` uses an explicit id list that does not include `pi`, so nothing else needs to change. Relaxing `KnownFields` would help future binaries but does nothing for the ones already out there.

**Closure condition:** either the published catalog carries no field unknown to released binaries, or the skew is accepted explicitly *and* a fetched-catalog parse failure emits a warning instead of degrading silently.

## `[blocker]` `internal/embedded/embedded.go:145`, `provider-catalog.yaml:298-299` — `liza-init-gate.ts` is a raw default-brand literal in a generated artifact

`PiInitGatePath` carefully brands the directory (`brand.RuntimeValues().GlobalDirName`) and then hardcodes the filename, so an `acme-agent` build deploys `~/.acme-agent/extensions/liza-init-gate.ts` and spawns `pi -p -e ~/.acme-agent/extensions/liza-init-gate.ts`. GUARDRAILS.md G1.3 (Tier 1) forbids raw default-brand identifiers in generated artifacts, advertised paths and distribution surfaces; none of ADR-0092's permitted categories cover a filename. The precedent one function away is `.opencode/tools/exec.ts` — unbranded.

`TestNonDefaultBrandBuildSmoke` only scans CLI help output, so CI passing is not evidence here.

**Why it matters beyond the guardrail:** the name is about to be locked into the published catalog. Renaming it afterwards desynchronizes catalog and binary — an old binary writing `liza-init-gate.ts` against a new catalog pointing at the new name gives pi a missing `-e` path, which by this PR's own account means pi refuses to start.

**Suggestion:** `extensions/init-gate.ts`. Four references: the `//go:embed` directive, `PiInitGatePath`, both catalog copies, and the `.ts` header comment.

**Closure condition:** no default-brand literal in the deployed filename or the catalog `run_args`.

---

## `[concern]` `internal/embedded/pi-extensions/liza-init-gate.ts` — the enforcement asset has no executable coverage

`pi_init_gate_test.go` greps the embedded bytes for substrings (`pi.on("tool_call"`, `fs.existsSync`, …). That proves deployment, not behavior. Nothing exercises the two contracts this file depends on: that pi's read tool passes its path as `event.input.path`, and that a returned `{block, reason}` is honored. If `input.path` is wrong the gate never records a read and blocks every mutating tool for the whole session; if the return contract is wrong the gate silently enforces nothing. The `.ts` is also type-loose (`new Set<string>()` on line 33, untyped params elsewhere) and nothing type-checks it.

The PR body reports a manual end-to-end pass on pi 0.84.4, which I'm taking as the evidence that it works today — the gap is regression coverage. A small node/bun test that imports the extension and feeds it synthetic `tool_call` events would cover both contracts.

**Closure condition:** either executable coverage of block-then-unblock, or an explicit note that the gate is manually verified per pi release.

## `[concern]` `support-docs/CONFIGURATION.md:1174-1190` — the CLI reference table omits `pi`

Every other provider, including the catalog-sourced `qwen` and `devin`, has a row naming its `setup`/`init` flags. `pi` has none, so the installed configuration reference does not mention the provider at all. `docs/provider-models.md` is a good doc but it is not that table, and it is not installed to `~/<global>/support-docs/`. Note the table uses `§BRAND_*§` macros — a `pi` row must too, and needs `make sync-embedded`.

---

## Suggestions

- `[suggestion]` `internal/agent/launch_resolver.go:428-434` — `brandGlobalDir` falls back to the bare dirname (`.liza`) when home cannot be resolved. The comment justifies it as "template rendering never fails", but the result is a relative `-e .liza/extensions/...` path that will not exist, converting a clear error into an obscure pi startup failure. An error here is more useful than a guaranteed-wrong path.
- `[suggestion]` `internal/embedded/pi-extensions/liza-init-gate.ts:11-12` — the header says the gate loads from `<projectRoot>/.pi/extensions/liza-init-gate.ts`. It is `{{globalDir}}/extensions/...`; the design note on `PiInitGatePath` explains why global was chosen, and this comment contradicts it.
- `[suggestion]` `internal/providers/catalog_test.go:147` — if `pi` stays in the published catalog, add it to the drift-guard id list. That test's own comment ("a change applied to one and not the other silently changes behavior depending on which catalog wins") applies most sharply to a provider whose `run_args` point at a file path. Moot if the first blocker is resolved by keeping `pi` embedded-only.
- `[suggestion]` `docs/README.md` — `docs/provider-models.md` is not in the index; every sibling doc is.

## `[appraisal]`

`missingDocs()` skipping absent docs and `sameFile()` comparing through `realpathSync` handle exactly the two cases that would otherwise bite — a pairing workspace with no `GUARDRAILS.md` deadlocking the session, and `/tmp` vs `/private/tmp` making a real read invisible. Both carry the reasoning in-line rather than in the PR.

---

**Approach:** Sound. Rebuilding on the provider catalog rather than reviving #74's `validCLIs` switch is the right call, and model selection reuses the existing `agent_tools` override instead of adding config surface — the first rung that holds.
**Absence:** Baseline taken from the PR's Changes table and test plan. Missing against it: the `CONFIGURATION.md` provider row and behavioral coverage of the gate.
**Net value:** *As submitted* — real pi support with enforcement parity, bought at the cost of freezing catalog delivery for the installed base and a brand leak in a shipped artifact. *With findings resolved* — clearly positive: a new backend on the existing architecture, no compat regression, no new config surface.
**Blast radius:** High — publishes to the live provider catalog, writes a global asset, and runs on the agent spawn path.
**Confidence:** High on the catalog and branding findings (reproduced locally against the recorded head); medium on the gate's runtime behavior, which CI does not execute.
**Sources:** full diff vs merge-base `e18231d3`; `internal/providers/catalog.go`, `internal/agent/launch_resolver.go`, `internal/commands/providers.go`, `internal/embedded/embedded.go`, all four new test files, `cmd/liza/brand_smoke_test.go`, `GUARDRAILS.md` G1.3, `support-docs/CONFIGURATION.md`; CI green on ubuntu/macos/windows for `16cd2e0f` (required check `license/cla` passing); one local probe parsing this branch's catalog with the pre-merge parser.
**Next step:** address the two blockers; the concerns and suggestions can ride along or route to follow-up.

LivioGama:
Addressed in `23f4dd82`, answering the [review](https://github.com/liza-mas/liza/pull/162#issuecomment-5782752296).

**Blockers**
- **`pi_extension` catalog skew**: `pi` is removed from the published `provider-catalog.yaml` and ships only in the embedded catalog.
  - Verified with the released parser (`main`): it parses the new catalog and rejects the old head's (`line 293: field pi_extension not found`).
  - `TestPublishedCatalogOmitsPiUntilReleased` keeps `pi` out of the published file until released binaries know the field.
  - The listing and detection paths read only the fetched catalog, so they still would not have seen `pi`. This covers `providers list`, `providers detect`, and setup auto-detect (`--yes` and the picker). They now go through `providers.WithEmbeddedBuiltins`, which skips only an embedded provider that collides with the fetched catalog.
  - Checked end to end: with the published catalog served locally and a fresh `HOME`, `list` and `detect` show `pi`, and `setup --yes` links `~/.pi/agent` and writes the gate.
- **Brand literal in the filename**: the gate is now `extensions/init-gate.ts` in the embed, `PiInitGatePath`, and both catalogs, and it is whitelisted in `.gitignore`.
  - The one remaining raw `liza setup` inside the gate is now `__BRAND_BINARY_NAME__ setup`.
  - `TestPiInitGateSourceHasNoRawBrandLiteral` guards this.

**Concerns**
- **Executable gate coverage**: `TestPiInitGateEnforcesInitSequence` and `TestPiInitGateSkipsAbsentDocs` run the rendered gate under bun, or node with type stripping, and feed it synthetic `tool_call` events.
  - They cover block then unblock, tilde, relative and symlinked read paths, and absent docs.
  - Mutating either contract (the `input.path` field or the `{block, reason}` return) fails them.
  - CI now sets up node 24 and fails instead of skipping when no runtime is found.
  - The gate is typed, and `tsgo --strict` is clean, including assignability to pi 0.87.1's `ExtensionAPI`.
- **`CONFIGURATION.md`**: added a `pi` row using `§BRAND_*§` macros.

**Suggestions**
- `{{globalDir}}` is left undefined when the home directory is unresolvable, so rendering fails with an explicit error instead of producing a relative `-e` path.
- The header comment is corrected.
- The drift-guard list item is moot, since `pi` is embedded-only.
- `docs/README.md` now indexes `provider-models.md`.
- Also fixed: the `setup --pi` help pointed at `~/.liza/agent/` instead of `~/.pi/agent/`.

`make lint`, `make test` (60 packages) and `make test-e2e` all pass on the final tree.

tangi-vass (approved):
**Review: Standard (Deep on the catalog + gate), round 2 — Approve** · head `23f4dd82`

Round 2: remaining 8→0, files 17→22

Blockers: None · Concerns: None · Suggestions: None

**Overall:** All eight round-1 findings are closed, and each fix is the smallest one that holds. The published catalog is now readable by released binaries, the deployed gate carries no brand literal, and the gate is covered by tests that actually run it.

## Reconciliation

| Round-1 finding | Status | Evidence at `23f4dd82` |
|---|---|---|
| `[blocker]` `pi_extension` in the published catalog | **RESOLVED** | Compared with base `e18231d3`, the only change to `provider-catalog.yaml` is `env_files: [codex.env]`, and the base parser already knows that field (`catalog.go:162`). `pi` resolves from the embedded catalog. `TestPublishedCatalogOmitsPiUntilReleased` keeps it that way. |
| `[blocker]` default-brand filename in the deployed artifact | **RESOLVED** | `extensions/init-gate.ts` is used in the embed, `PiInitGatePath` and the embedded `run_args`. The gate source uses `__BRAND_*__` macros throughout, and `TestPiInitGateSourceHasNoRawBrandLiteral` guards it. |
| `[concern]` no executable gate coverage | **RESOLVED** | `TestPiInitGateEnforcesInitSequence` and `TestPiInitGateSkipsAbsentDocs` run the rendered gate and send it synthetic `tool_call` events. They cover block then unblock, tilde/relative/symlinked read paths and absent docs. Breaking either the `input.path` contract or the `{block, reason}` contract fails them. CI provisions node 24, and `skipOrFailWithoutRuntime` fails under `CI` instead of skipping. |
| `[concern]` `CONFIGURATION.md` missing `pi` | **RESOLVED** | A row was added with `§BRAND_*§` macros. |
| `[suggestion]` `{{globalDir}}` relative fallback | **RESOLVED** | The variable is left undefined unless `paths.GlobalLizaDir()` returns an absolute path. Rendering then fails with a message that names the cause. |
| `[suggestion]` gate header location | **RESOLVED** | It now says `~/__BRAND_GLOBAL_DIRNAME__/extensions/init-gate.ts`. |
| `[suggestion]` drift-guard id list | **RESOLVED (moot)** | `pi` is embedded-only. |
| `[suggestion]` `docs/README.md` index | **RESOLVED** | The index now lists `provider-models.md`. |

**Growth (17→22 files) was declared and is justified.** Round 1 said keeping `pi` embedded-only needed no other change. That was wrong for `providers list`, `providers detect` and setup auto-detect, which read only the fetched catalog. `WithEmbeddedBuiltins` closes that gap. It appends an embedded provider only when neither its id nor its synthesized `-acp` id nor its alias collides, and it validates each candidate first. The published and embedded catalogs differ only in `pi`, so no other provider changes behavior. The other new files are the CI node step, the `.gitignore` whitelist for the renamed asset, and the two docs rows. The author's mapping comment declares every one of them.

**Consistency check:** the launch path (`paths.GlobalLizaDir()`) and the writer (`PiInitGatePath`) resolve the same directory. Both use `paths.UserHomeDir()` + the branded global dirname, so the `-e` path and the deployed gate cannot diverge.

**No new P0–P2 issues** in the corrective delta.

---

**Approach:** Settled in round 1.
**Absence:** Baseline is the round-1 findings plus the author's mapping comment. Nothing is missing.
**Net value:** *As submitted* — pi support with enforcement parity, no published-catalog skew, no brand leak, and the gate under regression coverage. *With findings resolved* — same; nothing remains.
**Blast radius:** High, but now contained: the published catalog change is limited to one field older binaries already know, and pi reaches users only through the binary.
**Confidence:** High. The catalog delta against base and the brand macros were checked directly against source. Gate behavior rests on the new executable tests running in CI for this head, plus the author's manual pi runs.
**Sources:** corrective diff `16cd2e0f..23f4dd82`, full diff vs merge-base `e18231d3`; `internal/embedded/embedded.go`, `pi-extensions/init-gate.ts`, `pi_init_gate_test.go`, `internal/providers/catalog.go`, `pi_catalog_test.go`, `internal/agent/launch_resolver.go`, `internal/paths/paths.go`, `internal/commands/providers.go`, `cmd/liza/cmd_init.go`, `provider_catalog_cmd_test.go`, `.github/workflows/ci.yml`; CI run 35926667616 green on ubuntu/macos/windows for `23f4dd82` (includes the executed gate tests under node 24); required check `license/cla` passing.
**Next step:** merge.

